PR #10, opened 2021-02-23
Css styling

kamara-moses:
I have added the styling to the task form.


index.html
```diff
@@ -42,6 +42,7 @@
       <div class="inputBox">
         <label for="status">Status:</label>
         <select id="status">
+          <option value="none">None</option>
```
swarajpure, line 45: There is going to be no status like `none`, we will not need it.
Instead we can have `blocked` for any task which is blocked because of some other task

swarajpure (commented):
Thank you this PR 🎉 

Mentioning one small change. 
Also, can you please add some screenshots? Since this is PR changing the UI, it'll be better for anyone to easily get what is being changed 😄

index.html
kamara-moses: I will changed none to blocked.

kamara-moses: I will add screenshots of the before and after to the readme.

swarajpure:
No longer needed. Refer #15

PR #11, opened 2021-02-23
changed the status of none to blocked

kamara-moses:
I have changed the status option of none to blocked. I also added the styling to the taskForm as you can see in the before and after screenshots.
after
before


swarajpure (changes requested):
Can you please remove the CSS and commit again? We just need the `blocked` status in this PR

ankushdharkar (approved):

PR #13, opened 2021-03-14
added blocked as a status for certain users who are blocked

kamara-moses:
There is a blocked status now on the task form. This can be used to block certain task that are not to be done until otherwise stated.


swarajpure (approved):
Thanks for this one! 🎉

PR #15, opened 2021-03-15
added styling to the taskform and maintaining the same asthetic as th…

kamara-moses:
I added some styling to the taskForm. I looked at the how the other forms are styled and maintained the same aesthetic as not to mess up how the other forms are handle or styled.

Here is the before the styling:
Screen Shot 2021-03-14 at 11 48 24 PM

Here is the after styling:
Screen Shot 2021-03-14 at 11 49 11 PM

### DIFF
--- a/index.html
+++ b/index.html
@@ -22,10 +22,16 @@
         <input type="text" id="featureUrl" name="featureUrl" placeholder="Feature URL">
       </div>
       <div class="inputBox">
-        <label for="dev">Dev</label>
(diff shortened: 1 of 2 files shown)


swarajpure (approved):
Looks absolutely awesome! Great job @kamara-moses! 🔥 

I'm approving from my side. Let @whyDontI approve. 

Please don't merge till then.

whyDontI (approved):
LGTM 👍🏽

Issue #21, opened 2021-04-08
Refactor task form


whyDontI:
## To Do
- Mark required fields with `*`
- Change `Gold, Silver, Bronze` to `Dinero & Neelam`
- Prefill default values
  - Dinero - 100
  - Neelam - 0
  - Percent completed - 0
  - Started - today's date
  - Status - pending
  - Preselect `Dev` in radio buttons `Dev / Group`
- Keep submit button disabled till all the required fields are filled without any errors

bhk31:
I will be working on this.

bhk31:
@whyDontI I have Few questions.

- Which fields are required in form or all are required ?
- You said Gold, Silver, Bronze to Dinero & Neelam So for Gold => Dinero, Silver => Neelam What about Bronze? Or Do we need to keep two only Dinero & Neelam.
-

whyDontI:
1) Required fields
- title
- purpose
- type
- startedOn
- status
- ownerId 

2) Only Dinero and Neelam

PR #23, opened 2021-04-10
Refactor task form #21

bhk31:
Before 
before

After
after

### DIFF
--- a/index.html
+++ b/index.html
@@ -16,8 +16,8 @@
         <input required class="input" type="text" name="title" placeholder="Title" id="title">
       </div>
       <div class="inputBox">
-        <label for="purpose">Purpose:<span class="required">*</span></label>
-        <input required class="input" type="text" id="purpose" name="purpose" placeholder="Purpose">
+        <label for="purpose">Purpose:</label>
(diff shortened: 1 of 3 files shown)


bhk31:
This PR also has changes related to issue [24](https://github.com/Real-Dev-Squad/website-dashboard/issues/24).

whyDontI (changes requested):
- @bhk31 Can you also make sure that the task data we are sending with POST request adheres to this [contract](https://github.com/Real-Dev-Squad/website-api-contracts/tree/main/tasks#post-tasks)
- Make sure the date we send to the backend is an `epoch / Unix timestamp`

whyDontI (changes requested):
Great work @bhk31,
Requesting some changes.
- Dates should be sent as epoch strings ( Right now they are being sent as numbers)
- Change `dev` to `feature`
- `assignedTo` should be `assignee`
- Send `dinero and neelam` in `completionAward and lossRate` fields
devtools-ss

whyDontI (approved):
Great work @bhk31 🥳

PR #50, opened 2021-06-25
formatted code with prettier cli

abhinav1590:


swarajpure:
Closed in lieu of #51

PR #54, opened 2021-06-26
formatted codes using prettierCLI

abhinav1590:


swarajpure (approved):
Looking great now!
Thanks for the fix. Hope you learnt about prettier CLI pretty well 😉

PR #60, opened 2021-06-29
github CI to check code format for every push 

abhinav1590:


.github/workflows/CI.yml
```diff
@@ -0,0 +1,13 @@
+name: CI
+on:
+push:
+branches:[ develop ]
+jobs:
+ runs-on: ubuntu-latest:
```
swarajpure, line 6: I don't think this will work. YAML works on indentation and it's wrong here.

If this works, you should see this (either passing or failing) on this PR:
image


Please refer this: https://github.com/Real-Dev-Squad/website-backend/blob/develop/.github/workflows/test.yml

swarajpure (changes requested):
Please check again, your file seems to be missing node version 🤔

.github/workflows/CI.yml
```diff
@@ -0,0 +1,24 @@
...
+
+    steps:
+      - uses: actions/checkout@v2
+      - name: Use Node.js ${{ matrix.node-version }}
+        uses: actions/setup-node@v1
+        with:
+          node-version: ${{ matrix.node-version }}
+      - run: npm install
+      - run: npm install --save-dev --save-exact prettier
+      - run: npx prettier --check .
```
swarajpure, line 24: Also, we don't need line 23 since we're already installing required packages on line 22
Also, on line 24, we just declare what needs to be run from the scripts, let's not create anything here. It should just be `npm run check` here. It's the package json's job to have defined what's `check` script
I hope that makes sense

PR #68, opened 2021-07-14
styling changes of submit button

ghosharnab969:
submit button earlier :
image

submit button later ( while disabled) : 
image

submit button later ( while enabled) : 
image

### DIFF
--- a/main.css
+++ b/main.css
@@ -81,8 +81,8 @@ body {
 }
 
 .disabled {
-  color: #fff;
-  background-color: #0000ee;
+  color: rgb(204 200 200);
+  background-color: #7171fd;
   opacity: 0.5;
 }
 
--- a/main.css
+++ b/main.css
@@ -76,16 +76,14 @@ body {
 }
 
 .active {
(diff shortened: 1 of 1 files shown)


main.css
```diff
@@ -76,16 +76,14 @@ body {
...
 }
 
 .disabled {
-  color: -internal-light-dark(rgba(16, 16, 16, 0.3), rgba(255, 255, 255, 0.3));
-  background-color: -internal-light-dark(
-    rgba(239, 239, 239, 0.3),
-    rgba(19, 1, 1, 0.3)
-  );
+  color: #fff;
+  background-color: #0000ee;
```
ankushdharkar, line 85: 'active' and 'disabled' classes have the same background color?

Also, are these keywords describing the same behaviour for the keywords outlined in [CSS Pseudo selectors](https://developer.mozilla.org/en-US/docs/Web/CSS/Pseudo-classes)?

ankushdharkar (changes requested):

PR #75, opened 2021-07-24
Change different placeholder font to similar 

anshika29:


swarajpure (changes requested):
How did package-lock change? Please revert the change in package-lock since you have not installed/removed/updated any package

You can google on how to do that :)

Also, please add before and after screenshots

anshika29:
*before
Screenshot (1112)_LI

*after
Screenshot (1110)_LI

swarajpure (approved):
Great!
Thanks for the fix! ⚡

PR #88, opened 2021-08-23
purpose box don't go outside the form horizontally

singh3069:
Fix #73

PR #91, opened 2021-08-26
Solved purpose box issue

oskorp:


swarajpure:
No activity since a month. Fixed in #88

Issue #92, opened 2021-08-26
Bug: footer not at the bottom on mainpage


swarajpure:
On https://dashboard.realdevsquad.com/, you'll see:
![image](https://user-images.githubusercontent.com/56217868/131016768-aa855f6d-f140-4408-8b5f-9ed0671b4c07.png)

The line `The contents of this website are deployed from this open sourced repo` is supposed to be a footer and it should be at the bottom of the page. Have a look here, how it should be: https://dashboard.realdevsquad.com/task/create.html

We could have a simple text like `Welcome to the dashboard site!` in the center of the screen as a heading maybe?

aman7870:
issue is fixed

PR #94, opened 2021-08-26
fix footer bug

shubhamdsm:
#92 
Before - 
131016768-aa855f6d-f140-4408-8b5f-9ed0671b4c07
After - 
fix-footer
After - Mobile View
chrome-capture

### DIFF
--- a/index.html
+++ b/index.html
@@ -1,32 +1,33 @@
 <!DOCTYPE html>
 <html lang="en">
+  <head>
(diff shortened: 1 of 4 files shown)


aman7870 (changes requested):
Thanks for PR!

Requesting changes.

Please fix the footer

Screenshot from 2021-09-09 07-52-14

aman7870:
issue is fixed

Issue #102, opened 2021-11-12
Bug fixes for task creation form


Pavangbhat:
**We have a few bugs to be fixed** :point_down: 
- [ ] Purpose field should be optional
- [ ] Feature URL can be empty
- [ ] Links can be empty
- [ ] Assignee field should not be empty
- [ ] User should not be able to set a value less than 0 for these fields **Percent Completed, Completion Award Dinero,Neelam, Neelam**
- [ ] Clear all the user entered data after successful creation of a task

**References** :point_down:

https://user-images.githubusercontent.com/45538392/141522251-61a51534-91fd-4e2c-9185-a9b39eee2864.mp4

**API-Contracts:** https://github.com/Real-Dev-Squad/website-api-contracts/tree/main/tasks

PR #105, opened 2021-11-14
fixed bugs for dashboard site

Pavangbhat:
**Bugs fixed** :point_down: 

- [x] Purpose field should be optional
- [x] Feature URL can be empty
- [x] Links can be empty
- [x] Participants can be empty
- [x] Assignee field can be empty
- [x] User should not be able to set a value less than 0 for these fields **Percent Completed, Completion Award Dinero, Neelam, Neelam**
- [x] Clear all the user entered data after successful creation of a task

closes : #102

### DIFF
--- a/main.css
+++ b/main.css
@@ -42,6 +42,9 @@ body {
 .inputBox label {
   font-size: 1rem;
   font-weight: 700;
+  width: 100%;
+  text-align: left;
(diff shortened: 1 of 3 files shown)


lakshayman (approved):
Everything looks good to me

PR #115, opened 2021-11-25
Enhanced ends-on section

ojas-elawadhi:
- [x] Default ends on date should be 7 days
 - [x] Move number of days remaining on Ends on line for eg: Ends on:* 2021-11-26 7 days remaining
 - [x] Once date has been set date picker, input box must be hidden if date is valid
 - [x] Remove all commented lines in setEndDate function
 - [x] Make 5px Clickable Region Around Edit

### DIFF
--- a/main.css
+++ b/main.css
@@ -207,8 +207,8 @@ footer {
 #remainingDays {
   /* all:unset; */
   float: left;
-  margin-block: -0.5em 0.75em;
-  padding-left: 0.5em;
+  margin-block: -3.9em 0.75em;
+  padding-left: 12.5em;
   font-size: 0.875em;
 }
 
@@ -221,11 +221,11 @@ footer {
   height: fit-content;
   font-size: 0.625em;
   opacity: 60%;
-  margin-block: 0.25em;
   right: 5px;
   cursor: pointer;
   display: block;
   transition: opacity ease 0.25s;
+  padding: 5px 10px 5px 0px;
 }
 
 .inputBox .preview {
--- a/main.css
+++ b/main.css
@@ -206,10 +206,11 @@ footer {
 
 #remainingDays {
   /* all:unset; */
-  float: left;
-  margin-block: -3.9em 0.75em;
-  padding-left: 12.5em;
+  float: right;
+  margin-block: 0.1em 0.75em;
+  padding-right: 19.5em;
   font-size: 0.875em;
+  font-weight: 400;
 }
 
 .notEditing {
--- a/main.css
+++ b/main.css
@@ -206,11 +206,10 @@ footer {
 
 #remainingDays {
   /* all:unset; */
-  float: right;
-  margin-block: 0.1em 0.75em;
-  padding-right: 19.5em;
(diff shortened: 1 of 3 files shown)


main.css
```diff
@@ -207,8 +207,8 @@ footer {
 #remainingDays {
   /* all:unset; */
   float: left;
-  margin-block: -0.5em 0.75em;
-  padding-left: 0.5em;
+  margin-block: -3.9em 0.75em;
+  padding-left: 12.5em;
```
lakshayman, line 211: Why are you giving margins to the remaining days? Can't you place the remaining days in the label?

lakshayman (changes requested):
Looks good to me

script.js
```diff
@@ -288,3 +282,13 @@ edits.forEach((edit, index) => {
   });
   edit.append(element);
 });
+
+function handleDateChange(event) {
+  const preview = event.target.previousElementSibling.children[1];
```
lakshayman, line 287: Place const input = event.target in starting of the function
and make it

```suggestion
  const preview = input.previousElementSibling.children[1];
```

script.js
```diff
@@ -288,3 +282,13 @@ edits.forEach((edit, index) => {
...
+  const edit = event.target.previousElementSibling.children[2];
```
lakshayman, line 288: This one too
```suggestion
  const edit = input.previousElementSibling.children[2];
```

Pavangbhat (changes requested):
image
- [ ] By default ends on should be 7 days
image
- [ ] We need to have some margin here

main.css
ojas-elawadhi: Yes I can place it in label but still I will need to give margins and change some properties to it to place it to correct position.
Please tell which is more correct way? I will do it that way...

script.js
```diff
@@ -45,28 +45,24 @@ const isNoteworthy = document.getElementById('isNoteworthy');
...
     endDate.parentElement.querySelector('em').innerHTML = `${
       endTimeArray[2]
     }-${endTimeArray[0].padStart(2, '0')}-${endTimeArray[1].padStart(2, '0')}`;
 
-  const remainingDays = document.getElementById('remainingDays').children[0];
-  remainingDays.innerHTML = Math.ceil(
-    (endTime - startTime) / (1000 * 60 * 60 * 24),
-  );
-};
+/*   const remainingDays = document.getElementById('remainingDays').children[0];*/
```
Pavangbhat, line 61: Please remove the comments

Pavangbhat (commented):
Tested on firefox  looks like it needs some margin 
image

script.js
ojas-elawadhi: Ok will do the changes....but now it is showing correct default remaining days (7) on your side?

Pavangbhat: Yes Ojas

Pavangbhat (changes requested):
Please make it responsive as well

image
Other than that everything looks good to me

PR #127, opened 2021-12-25
fix

vinayak-trivedi:
Made text non selectable.


Pavangbhat (changes requested):
image
It's saying that code style is not following some rules so **npm run check** fails that is why you have to run **npm run fix** which fixes the code style then commit and push those changes. You have run it in your vs code terminal  
One suggestion:- 

- Use proper tile for PRs and commit messages (Please refer to closed PRs you get an idea)

PR #144, opened 2022-02-07
Fixed auto clear up of filled information if task creation fails

DeRaowl:
- Fixed auto-clear up of filled information if task creation fails
https://user-images.githubusercontent.com/26904087/152802423-e9c9db05-a750-45b5-b6e4-947ea43f4d98.mp4

### DIFF
--- a/src/task/script.js
+++ b/src/task/script.js
@@ -206,8 +206,10 @@ taskForm.onsubmit = async (e) => {
 
     const result = await response.json();
 
+    if (response.status>=200 && response.status<400) {
+      window.location.reload(true);
+    }
     alert(result.message);
-    window.location.reload(true);
   } catch (error) {
     alert(`Error: ${error}`);
   } finally {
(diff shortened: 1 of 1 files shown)


src/task/script.js
```diff
@@ -206,8 +206,10 @@ taskForm.onsubmit = async (e) => {
 
     const result = await response.json();
 
+    if (response.ok) {
+      window.location.reload(true);
+    }
     alert(result.message);
```
Pavangbhat, line 212: What if the task was created successfully?
Will the user get a response message shown?

Pavangbhat (changes requested):

src/task/script.js
DeRaowl: Yes, Only reload has been avoided. Users receive messages after successful task creation.

Pavangbhat: No, I don't think so 
Here line 210 will be executed if the response is successful which reloads the page so line 212 will not be shown to the user.
Please fix it

Pavangbhat (approved):

Pavangbhat (changes requested):
Some of the checks are failing 
Please check why they are failing and push the changes

Issue #177, opened 2022-07-08
Issue in Assignee field | Autocomplete search


DeRaowl:
Currently when we are entering any name it is listing all the members of RDS. Can we create scroll bar so that it will only take fixed width for auto complete.

Snippets for reference:
![image](https://user-images.githubusercontent.com/26904087/178026120-01c1d6be-7e73-4eea-ab13-b3e6bce36a06.png)
![image](https://user-images.githubusercontent.com/26904087/178026177-c69e9ab8-db11-4692-a35a-d9b61ee63086.png)

rajat-mehra05:
Can you assign me this task ? Thanks.

DeRaowl:
Assigned to you

PR #181, opened 2022-07-19
fix: added scrollbar to the autocomplete-search

rajat-mehra05:
Closes: Issue No. #177 

image

Here's the demo:

[rds.webm](https://user-images.githubusercontent.com/35677839/179937999-65ac62ef-927d-4067-b1db-b41d8a3d42d1.webm)

### DIFF
--- a/src/task/style.css
+++ b/src/task/style.css
@@ -220,6 +220,9 @@ footer {
 
 #list-items {
   width: 100%;
+  max-height: 12.5rem;
+  overflow-y: scroll;
+  overflow-x: hidden;
   margin-top: 2rem;
   position: absolute;
   border: 1px solid #d4d4d4;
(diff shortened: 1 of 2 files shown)


src/task/style.css
```diff
@@ -242,6 +242,11 @@ footer {
   margin: 0;
 }
 
+#list-items {
+  max-height: 200px;
+  overflow-y: scroll;
```
DeRaowl, line 247: What about overflow-x: If I change css property in existing codebase I am seeing scroll in x axis. Can you check

src/task/style.css
```diff
@@ -242,6 +242,11 @@ footer {
   margin: 0;
 }
 
+#list-items {
```
DeRaowl, line 245: We already have styles for #list-items can you check regarding this?

src/task/style.css
```diff
@@ -242,6 +242,11 @@ footer {
...
+  max-height: 200px;
```
DeRaowl, line 246: Can we make use of relative units like rem

src/task/style.css
rajat-mehra05: Resolved.

src/task/style.css
DeRaowl: LGTM

Pavangbhat (approved):

PR #188, opened 2022-07-24
Added fields for task levels

### LINKED ISSUE #189: Add two new fields in the `Create Task` form
Add the following data fields required for the Leveling of Tasks
- Add a category field
- Add a level field

For reference : https://github.com/Real-Dev-Squad/todo-action-items/issues/72

rohan09-raj:
Closes #189


src/task/index.html
```diff
@@ -106,6 +106,38 @@
...
+              Category:<span class="required">*</span>
+            </label>
+            <select
+              required
+              class="input notEditing"
+              id="category"
+              name="category"
+            >
+              <option selected value="system_design">System Design</option>
+              <option selected value="frontend">Frontend</option>
```
Mir-SA, line 123: Why are both fields selected?
Also this dropdown currently have only 4 fields, how will we add `other` skill? Suppose a custom skill like `React`

src/task/style.css
```diff
@@ -104,6 +104,8 @@ body {
   font-size: medium;
 }
 
+#category,
+#level,
 #status {
```
shubham-y, line 109: Isn't it better to have a class for it?

rohan09-raj: The class is already there and it is applied to many more elements.
That class is used in JS for a separate purpose
The code needs refactoring and I guess @Pavangbhat is already taking care of it separately

shubham-y (approved):

Pavangbhat (dismissed):

src/task/index.html
```diff
@@ -106,6 +106,37 @@
...
+            </label>
+            <select
+              required
+              class="input notEditing"
+              id="category"
+              name="category"
+            >
+              <option selected value="system_design">System Design</option>
+              <option value="frontend">Frontend</option>
+              <option value="backend">Backend</option>
```
vinayak-trivedi, line 124: can we make all the letters uppercase in the category to maintain consistency?

rohan09-raj: The labels are consistent.
We have the same in status also
image

rohan09-raj: Are you concerned about the values ??

vinayak-trivedi: yes, all the values of status are also stored in uppercase in our DB

vinayak-trivedi (approved):
Thanks for doing the changes, LGTM 👍

PR #193, opened 2022-08-14
adds image of selected user

### LINKED ISSUE #185: Show image of selected user
On right of the input box
image

poonammohata:
Resolves #185 

This is how changes are looking

https://user-images.githubusercontent.com/54801934/184539754-e6809510-dfeb-40f9-a440-26523c8aa6bd.mov

This is how it looks with new changes - added spinner and showing round image @ankushdharkar @Pavangbhat 

https://user-images.githubusercontent.com/54801934/185784992-1cf01140-1468-45ef-a038-57b2f916cbf4.mov


Pavangbhat (approved):
Great work, Looks good to me
Some changes might be needed

Pavangbhat:
I think a bit of border radius would look good  for the image

src/task/script.js
```diff
@@ -387,20 +388,30 @@ function clearUserNotFound() {
 function createSuggestionsList(matches) {
   const listItems = document.getElementById('list-items');
   if (matches.length) {
-    matches.map(({ username }) => {
+    matches.map(({ username, picture = {} }) => {
+      const defaultImageUrl =
+        'https://members.realdevsquad.com/images/Avatar.png?90.2370159455815';
```
Pavangbhat, line 393: Will the URL always remain the same?
It would good to create a images folder at /src and place this image there

Pavangbhat (changes requested):

Maheima:
LGTM

src/task/script.js
poonammohata: yes, the default avatar will be going to be the same, I thought of putting the image in a folder but wasn't sure which avatar image I should be using here, so, for now, the one is getting used on the member's page, I used it here. Should I put this one in an image folder, or do I have to use a different avatar?

Pavangbhat: This avatar looks good.
What if for some reason URL is changed,  the image gets updated, members site goes down, etc I think it would be better if we place this image in the dashboard site repo itself since this will remove our dependency on the member's site

poonammohata: make sense, will place the image in a folder

src/task/style.css
```diff
@@ -258,3 +258,10 @@ footer {
   color: rgb(255, 92, 92);
   font-size: 0.75rem;
 }
+
+#assignee {
+  background-repeat: no-repeat;
+  background-position: right;
+  background-size: 42px;
```
shubham-zebu, line 265: Can we use the shorthand [background](https://developer.mozilla.org/en-US/docs/Web/CSS/background) CSS property for all these?

shubham-y, line 265: Can we use the shorthand [background](https://github.com/Real-Dev-Squad/website-dashboard/pull/193/files#diff-5a43e4e279e445a814167ecbef27ccf07a7f5ca55283e18504b81279b0f39270R263-R265) CSS property for all these?

src/task/style.css
```diff
@@ -258,3 +258,24 @@ footer {
...
+  display: none;
+}
+
+#assigneeAvatarLoader {
+  position: absolute;
+  top: 41%;
+  right: 3%;
+  display: none;
+  border: 3px solid #f3f3f3;
+  border-top: 3px solid #1d58d8;
```
ankushdharkar, line 280: Nit: Since the border width and border style is the same for the top border from the line above, you could just set the top border color using[ border-top-color](https://developer.mozilla.org/en-US/docs/Web/CSS/border-top-color) CSS property

src/task/style.css
poonammohata: @shubham-y in new changes I am using background-position and background-size property only, and this time I used shorthand property as suggested by you

src/task/style.css
```diff
@@ -258,3 +258,23 @@ footer {
   color: rgb(255, 92, 92);
   font-size: 0.75rem;
 }
+
+#assigneeAvatar {
+  background: right / 42px;
```
Pavangbhat, line 263: Does this need to be wrapped around **calc()** ?

Pavangbhat (commented):
Hi @poonammohata 
Had one small doubt here https://github.com/Real-Dev-Squad/website-dashboard/pull/193#discussion_r953814318

src/task/style.css
poonammohata: no, actually I added background-position and background-size property in this, and this is how the syntax is, if we add position and size, have to add slash ( / ), otherwise it doesn't work

Pavangbhat: That was help, Thank you

Pavangbhat (approved):
Thank you @poonammohata 
Looks good to me

src/task/script.js
```diff
@@ -389,20 +393,32 @@ function clearUserNotFound() {
 function createSuggestionsList(matches) {
   const listItems = document.getElementById('list-items');
   if (matches.length) {
-    matches.map(({ username }) => {
+    assigneeAvatarLoader.style.display = 'initial';
```
Neha, line 396: Prefer having class name changes over inline style. Controlling style via JavaScript is an expensive task.

src/task/script.js
```diff
@@ -389,20 +393,32 @@ function clearUserNotFound() {
...
+    matches.map(({ username, picture = {} }) => {
+      const defaultImagePath = '../images/avatar.png';
```
Neha, line 398: This should move to a separate constant file to save the file name.

src/task/script.js
```diff
@@ -389,20 +393,32 @@ function clearUserNotFound() {
...
       listItems.appendChild(listItem);
     });
   }
 }
 
-function setAssignee(assignee) {
+function setAssignee(assignee, imgUrl) {
   assigneeEl.value = assignee;
   wasAssigneeSet = true;
+  assigneeAvatar.style.display = 'initial';
```
Neha, line 419: Prefer having class name changes over inline style. Controlling style via JavaScript is an expensive task.

src/task/script.js
```diff
@@ -389,20 +393,32 @@ function clearUserNotFound() {
...
+  assigneeAvatarLoader.style.display = 'none';
```
Neha, line 420: I see there is the reusable task:

1. Hide/show of the elements or content
2. Update of the image
3. Creating element

Move all these to a reusable method  rather than re-creating the same code again and again

src/task/style.css
```diff
@@ -258,3 +258,23 @@ footer {
...
+  position: absolute;
+  top: 33%;
+  width: 40px;
+  height: 40px;
+  right: 3%;
+  border-radius: 50%;
+  display: none;
```
Neha, line 270: A better approach would be to move the `display:none` to a separate reusable class and keep the style of the avtaar as it is.

Neha (commented):
This project requires cleaning. There are a lot of anti-patterns I can see. At a longer run this code would become a struggle

poonammohata:
Thanks, @Neha  for reviewing this so thoroughly, should I make all these suggested changes in this PR only, or work on these, in the refactoring issues which you have created?

Neha:
I would suggest making the changes in the PR only. I have raised the issues for over all code base.

src/task/script.js
poonammohata: I have created a constant.js file and added this path, now as I have to use that constant.js file in this script.js file, I have to use modules, as now the script.js file is importing constants.js file, it also became module type, and now all the functions which are in the script.js file is not globally accessible, and that's why I am getting reference error on all the functions which are attached in tasks.html file, there is one solution to this issue is to attach all script.js file functions to a window object, but I am not sure this is the right way to go. 
Can u suggest what should be done in this case @Neha

src/task/script.js
```diff
@@ -1,4 +1,6 @@
 const API_BASE_URL = window.API_BASE_URL;
+const assigneeAvatar = document.getElementById('assigneeAvatar');
```
Neha, line 2: These should also move to constant. So, all the hardcode values should be part of the Constant.

src/task/script.js
```diff
@@ -389,20 +393,31 @@ function clearUserNotFound() {
...
-    matches.map(({ username }) => {
+    addDisplay(assigneeAvatarLoader);
+    matches.map(({ username, picture = {} }) => {
+      const imageUrl =
+        picture.hasOwnProperty('url') && picture.url
+          ? picture.url
+          : DEFAULT_ASSIGNEE_IMAGE_PATH;
       const listItem = document.createElement('p');
       listItem.classList.add('list-item');
       listItem.style.cursor = 'pointer';
```
Neha, line 404: This style can't be merged with `list-item` class?

src/task/script.js
```diff
@@ -389,20 +393,31 @@ function clearUserNotFound() {
...
       listItems.appendChild(listItem);
     });
   }
 }
 
-function setAssignee(assignee) {
+function setAssignee(assignee, imgUrl) {
   assigneeEl.value = assignee;
   wasAssigneeSet = true;
+  addDisplay(assigneeAvatar);
```
Neha, line 418: The name of the function should be `addClass` and `removeClass` to make it more generic.

src/task/script.js
```diff
@@ -389,20 +393,31 @@ function clearUserNotFound() {
...
   const listItems = document.getElementById('list-items');
   if (matches.length) {
-    matches.map(({ username }) => {
+    addDisplay(assigneeAvatarLoader);
+    matches.map(({ username, picture = {} }) => {
+      const imageUrl =
+        picture.hasOwnProperty('url') && picture.url
+          ? picture.url
+          : DEFAULT_ASSIGNEE_IMAGE_PATH;
       const listItem = document.createElement('p');
```
Neha, line 402: Creating HTML elements should also move to reusable code. Where it should accept - element name, attributes, class, container.

element name and container should be mandatory. Rest can be optional

src/task/script.js
```diff
@@ -418,3 +433,11 @@ assigneeEl.addEventListener(
   'input',
   debounce((event) => fetchMembers(event.target.value), 500),
 );
+
+function addDisplay(element) {
```
Neha, line 437: Change the name to `addClassName` to make it more generic.

Move these functions to the utils.js file and export from there

src/task/script.js
```diff
@@ -418,3 +433,11 @@ assigneeEl.addEventListener(
...
+  element.classList.remove('element-display-remove');
+}
+
+function removeDisplay(element) {
```
Neha, line 441: Change the name to `removeClassName` to make it more generic.

Move these functions to the utils.js file and export from there

Neha (commented):
A few small comments I added. Please check

src/task/script.js
poonammohata: moved API_BASE_URL to constants, assigneeAvatar can also be moved to constants, but as you suggested for different types of constants we should have different constants files, and now we are working on the restructuring of whole code, so once that is done, will move the HTML element also to constant

src/task/script.js
poonammohata: merged this style with list-item class

src/task/script.js
poonammohata: updated the function name, as someone is already working on code restructure so once the code gets restructured will add this function to the utils file

src/task/script.js
poonammohata: updated the function name, as someone is already working on code restructure so once the code gets restructured will add this function to the utils file

src/task/script.js
poonammohata: updated the function name

src/task/script.js
poonammohata: This needs to be changed in the whole code, so once the code gets restructured, will create a function for this in util.js file and update it in the whole codebase

JC-Coder:
@pankajjs @yesyash 
This is implemented already. kindly close

Issue #200, opened 2022-08-21
Color highlight feature for profile diff data 


vin18:
### Issue
- Add a color highlight feature similar to Github that shows users' profile data differences in red and green color. 

### Description
- Currently, we are showing old data and new data as profile differences on the profile page. The UX can be improved by highlighting the old data with red background color and newly changed data in the green background color. This feature would be similar to Github's commit diff feature. 
- The data field which is not modified will not have any background color.

vin18:
I am taking up this issue.

RitikJaiswal75:
Vinit this issue could use a little more description

vin18:
> Vinit this issue could use a little more description

Updated. Thank you for the feedback,

PR #201, opened 2022-08-21
Add color highlight to compare old and new profile data on the profile page

vin18:
### Issue
Link to the ticket: : [Color highlight feature - Issue#200](https://github.com/Real-Dev-Squad/website-dashboard/issues/200)

### Solution Implemented
- Created old and new mock data of the user's profile since the only superuser has access to that data (Mock data discarded before committing)
- Compared each key-value pair of old and new data, converted it to string and trimmed it.
- Checked if values are not same, added the 'oldDiff' class (red background) and 'newDiff' class (green background) for the dissimilar value 

### Screenshot of the UI 
profile-diff-color-feature

### Note: 
- Since only the superuser has access to the data, the solution has been implemented with the mock data. This feature needs to be tested for real data.


src/profile/style.css
```diff
@@ -157,6 +157,19 @@ nav h1::before {
   border-right: 1px solid #c4c4c4;
 }
 
+.oldDiff {
+  background-color: #fd5757;
```
RitikJaiswal75, line 161: Please Use variables instead of directly assigning to hex

src/profile/style.css
```diff
@@ -157,6 +157,19 @@ nav h1::before {
...
+  padding: 4px;
+  border-radius: 2px;
+  color: #fff;
```
RitikJaiswal75, line 164: There is already a variable for this please use that

src/profile/utils.js
```diff
@@ -44,15 +44,29 @@ function getDataItem(data, itemName) {
...
+    const isValueEqual = String(oldValue).trim() === String(newValue).trim();
+
+    let diffClass;
+    if (!isValueEqual) {
+      diffClass = isOldData ? 'oldDiff' : 'newDiff';
+    }
+
+    const innerHTML = `
+      <span>${formatPropertyField(listItem)}:</span>
+      <span class=${diffClass}>${getDataItem(primaryData, listItem)}</span>
```
RitikJaiswal75, line 64: Can you try using js create element method

RitikJaiswal75 (changes requested):

RitikJaiswal75:
Some of your checks are failing, Lint checks.

src/profile/utils.js
```diff
@@ -224,6 +238,10 @@ function createCardComponent({ className, tagName, innerText, parent }) {
     component.innerText = innerText;
   }
 
+  if (innerHTML) {
+    component.innerHTML = innerHTML;
```
ankushdharkar, line 242: Are we potentially overriding the `innerText` with `innerHTML`?

vin18: Yes. Is that a problem? Because currently, we are rendering the key-value pair as a single text. E.g: `First name: Vinit`
I have refactored it and wrapped key and value in span elements and conditionally applied red and green background classes for the diff key-value pairs.

Is this okay or can it be improved with a better solution?

src/profile/utils.js
```diff
@@ -44,15 +44,29 @@ function getDataItem(data, itemName) {
...
+  const isNewData = prevSibling.toLowerCase().includes('new');
+
+  for (const listItem in primaryData) {
+    const oldValue = getDataItem(primaryData, listItem);
+    const newValue = getDataItem(secondaryData, listItem);
+    const isValueEqual = String(oldValue).trim() === String(newValue).trim();
+
+    let diffClass;
+    if (!isValueEqual) {
+      diffClass = isOldData ? 'oldDiff' : 'newDiff';
```
heyrandhir, line 59: we can create constants for 'oldDiff' and 'newDiff' strings.

vin18: Can you please check now?

src/profile/style.css
vin18: Can you please check now?

src/profile/style.css
vin18: Can you please check now?

src/profile/utils.js
vin18: Refactored the same with createElement syntax. Please check it.

src/profile/utils.js
vin18: Refactored these changes. Please check it. @ankushdharkar

src/profile/utils.js
```diff
@@ -44,17 +46,39 @@ function getDataItem(data, itemName) {
...
+      diffClass = isOldData ? OLD_DIFF_CLASS : NEW_DIFF_CLASS;
+    }
+
+    const fragment = new DocumentFragment();
+
+    let spanKey = document.createElement('span');
+    spanKey.textContent = `${formatPropertyField(listItem)}: `;
+    fragment.appendChild(spanKey);
+
+    let spanValue = document.createElement('span');
```
Pavangbhat, line 70: let's make use of **const** where ever and all possible

Pavangbhat (changes requested):
Great work @vin18 👏 
Requesting a small change

src/profile/utils.js
vin18: Thank you for your kind words @Pavangbhat. Can you please check it now?

RitikJaiswal75 (approved):
LGTM

RitikJaiswal75 (dismissed):

src/profile/style.css
```diff
@@ -157,6 +163,19 @@ nav h1::before {
...
+  border-radius: 2px;
+  color: var(--color-white);
+}
+
+.newDiff {
+  background-color: var(--color-green);
+  padding: 2px 4px;
+  border-radius: 2px;
+  color: var(--color-white);
+}
```
poonammohata, line 178: there are 3 properties i.e padding(left and right), border-radius, and color are the same in both oldDiff and newDiff classes. Can we have a separate reusable class that has these properties, and oldDiff and newDiff classes have only the properties which are different?

Pavangbhat (changes requested):
Thanks, @poonammohata for the review
There is one open comment please have a look

src/profile/style.css
vin18: Thanks, @poonammohata for your review. I have done the changes. Kindly review.

Pavangbhat (approved):

PR #209, opened 2022-09-08
Sync main with develop 

Pavangbhat:


src/profile/utils.js
```diff
@@ -104,13 +141,15 @@ function createCard({ oldData, newData, userId, username, profileDiffId }) {
...
+  const diffValues = checkDifferentValues(oldData, newData);
+
   const oldUserInfoList = createCardComponent({
     tagName: 'ul',
     className: 'userInfoListContainer',
     parent: oldDataContainer,
   });
 
-  displayList(oldData, oldUserInfoList);
+  displayList(oldData, oldUserInfoList, diffValues, OLD_DATA);
```
rajakvk, line 152: NIT: Pass parameters in an object like displayList(options);

src/profile/utils.js
```diff
@@ -130,7 +169,7 @@ function createCard({ oldData, newData, userId, username, profileDiffId }) {
     parent: newDataContainer,
   });
 
-  displayList(newData, newUserInfoList);
+  displayList(newData, newUserInfoList, diffValues, NEW_DATA);
```
rajakvk, line 172: NIT: Pass parameters in an object like displayList(options);

src/profile/utils.js
Pavangbhat: Thank you @rajakvk 
@vin18 let's consider while refactoring

Pavangbhat: cc @rohan09-raj

rohan09-raj (approved):

src/profile/utils.js
vin18: Got it, @Pavangbhat. Will handle this in global refactor PR. 
Thank you @rajakvk for the review.

PR #219, opened 2022-09-30
Sync dev to main

prakashchoudhary07:


src/online-members/sse-events.js
```diff
@@ -0,0 +1,70 @@
...
+  const eventSource = new EventSource(url, { withCredentials: true });
+
+  eventSource.onopen = (e) => {
+    console.info('The connection has been established.');
+  };
+
+  eventSource.onmessage = function (event) {
+    try {
+      const objectData = JSON.parse(event.data);
+      currentOnlineList = objectData.users;
```
harshith-venkatesh, line 18: @prakashchoudhary07 , is this causing this issue where we are looping through null data?

PR #221, opened 2022-10-01
Fix: tasks/null api call when clicked on users

prakashchoudhary07:
On clicking members to get task data, an API call with null as the user name was being made, That is being fixed in this

### DIFF
--- a/src/online-members/script.js
+++ b/src/online-members/script.js
@@ -138,7 +138,7 @@ async function generateMemberTaskData(username) {
 
   if (!username) {
     throw new Error(
-      'Some error occurred while fetching user data, please try again or contact admin',
+      'Some error occurred while getting user data, please try again or contact admin',
     );
   }
 
--- a/src/online-members/script.js
+++ b/src/online-members/script.js
@@ -99,10 +99,10 @@ function getMembersListContent(members, classList = []) {
   return ul;
 }
 
-function getTaskDataContent(tasks, classList = []) {
+function getTaskDataContent({ tasks, username, classList = [] }) {
   const div = createElement({ type: 'div' });
   const h3 = createElement({ type: 'h3' });
-  h3.appendChild(createTextNode('TASKS'));
+  h3.appendChild(createTextNode(`TASKS - ${username}`));
   const fragment = new DocumentFragment();
   tasks.forEach((task) => {
     const div = createElement({ type: 'div', classList: TASKS_CLASS_LIST });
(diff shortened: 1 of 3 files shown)


src/online-members/script.js
```diff
@@ -136,6 +136,12 @@ async function generateMemberTaskData(username) {
     return;
   }
 
+  if (!username) {
+    throw new Error(
+      'Some error occurred while getting user data, please try again or contact admin',
+    );
+  }
+
```
poonammohata, line 144: You have handled that if the username is the null in the "addEventListenerToMembersList" function already before calling the "getTaskDataContent" function there, then why are you checking this again here?

prakashchoudhary07: Yes, right this has been already handled, will remove it.
Thanks for pointing it out :smiley:

src/online-members/script.js
```diff
@@ -162,13 +162,16 @@ async function generateMembersList() {
...
 function addEventListenerToMembersList() {
-  const membersList = document.querySelector(`#${MEMBERS_CONTAINER_ID} > ul`);
+  const membersList = document.querySelector(`#${MEMBERS_CONTAINER_ID} > UL`);
   membersList.addEventListener('click', (event) => {
-    const membersDiv = event.target.nodeName === 'DIV';
-    if (!membersDiv) {
+    const memberElement = event.target.closest(`.${MEMBERS_CLASS}`);
+    if (!memberElement) {
       return;
     }
```
Pavangbhat, line 170: Is there any chance of dataset being null or undefined?

Pavangbhat (commented):
Have added a comment, please have a look

src/online-members/script.js
prakashchoudhary07: Yes is element which does not have data set is selected then it will be null

src/online-members/script.js
```diff
@@ -162,13 +162,16 @@ async function generateMembersList() {
...
-    const username = event.target.dataset.username;
+    const username = memberElement.dataset.username;
```
Pavangbhat, line 171: Should we add null checks here for dataset ?

Pavangbhat (commented):
@prakashchoudhary07

src/online-members/script.js
prakashchoudhary07: Isn't that being handled in the next line? or am I missing something here?

Pavangbhat: I am wondering if it's possible for dataset to be null/undefined, in that case it would be something like memeberElement.undefined.username which would throw an error
Please correct me if I am worng in understanding this

prakashchoudhary07: Dataset won't be null, it is a property that is available for all elements in HTML, so if set will return the value, or else will return undefined if values are not set, Please let me know if I am missing something

Pavangbhat: Got it

Pavangbhat:
LGTM

Pavangbhat (approved):

Issue #203, opened 2022-08-24
Create the constant file


Neha:
Problem:

Currently, the project has all the content hardcoded in the code. This is a non-scalable and non-maintainable approach.

Solution:

Move the content to the constant file. There could be different constant files:

1. URLs, APIs, etc.
2. UI elements
3. Content

PR #240, opened 2022-10-24
adds user management page

### LINKED ISSUE #227: Table View and Tile View Screen for the Users on the User Management
On the dashboard Site, there will be a button called User Management that should only be visible to super users. On clicking that the User will be directed to the Listing Screen. This screen will list all the Users in the tile view or table view.  There should be a toggle button on top to switch between both modes. Also, a search bar at the top to search for specific users.

Note: The page should be responsive. We need to also write unit and integration tests and also take care of accessibility. 

Refer to this [figma link](https://www.figma.com/file/UsEeJHsLrJp9P0lUc5dDl1/UM-Service?node-id=0%3A1) for the design. You can tweak the design if you can make it look better.

poonammohata:
### Issue
Closes #227

### What is the change?
Added a button called 'User Management' that is only visible to super users. On clicking that the User will be directed to the Listing Screen. This screen will list all the Users in the tile view or table view. There is a toggle button on top to switch between both modes. Also, a search bar at the top to search for specific users.

### Is Development Tested?
- yes

### Short video of how this will look on the website

https://user-images.githubusercontent.com/54801934/197562890-3ea75143-9d49-4580-826e-fd72040d88fd.mp4


heyrandhir:
Hi @poonammohata ,
can we add more tiles in the left and right regions. There is quite a space available for the tiles. 
image
The same for the table view. we can add more list items on the left and right.

poonammohata:
> Hi @poonammohata , can we add more tiles in the left and right regions. There is quite a space available for the tiles. image The same for the table view. we can add more list items on the left and right.

@heyrandhir is it looking fine, or need some changes?
image
image

heyrandhir:
@poonammohata it looks good. Thanks for the changes.

heyrandhir (dismissed):
LGTM 👍🏻

src/user-management/constants.js
```diff
@@ -0,0 +1,7 @@
+const RDS_API_USERS = 'https://api.realdevsquad.com/users';
```
RitikJaiswal75, line 1: Cant we use API_BASE_URL+"/users/" to maintain DRY

RitikJaiswal75 (changes requested):
Also, please remove .DS_Store files

src/script.js
```diff
@@ -0,0 +1,8 @@
+async function showUserManagementButton() {
+  const isSuperUser = await checkUserIsSuperUser();
+  if (isSuperUser) {
+    userManagementBtn.classList.remove('element-display-remove');
+  }
+}
+
+showUserManagementButton();
```
RitikJaiswal75, line 8: Lets export this method so that it can be reused anywhere else as well

RitikJaiswal75 (changes requested):

src/constants.js
```diff
@@ -0,0 +1,2 @@
+const API_BASE_URL = 'https://api.realdevsquad.com';
+const userManagementBtn = document.getElementById('user-management-btn');
```
ankushdharkar, line 2: Constants should not contain a reference to a live DOM object. What if this object is removed using JS in future somewhere else. This reference will become orphaned and wrong.

src/index.html
```diff
@@ -27,6 +30,14 @@
       >
         Profile
       </button>
+      <button
+        id="user-management-btn"
+        class="element-display-remove"
+        type="button"
+        onclick="location.href='user-management/index.html'"
```
ankushdharkar, line 37: Why is this not an `<a>` tag?

src/user-management/index.html
```diff
@@ -0,0 +1,44 @@
...
+        </button>
+      </div>
+      <div id="user-list">
+        <div id="loader"></div>
+      </div>
+    </section>
+  </body>
+  <script src="constants.js"></script>
+  <script src="utils.js"></script>
+  <script src="script.js"></script>
```
ankushdharkar, line 43: Why do these scripts not use `defer` like earlier? We should try to maintain consistency in our usage patterns.

src/user-management/script.js
```diff
@@ -0,0 +1,127 @@
...
+  return (...args) => {
+    if (timerId) {
+      clearTimeout(timerId);
+    }
+
+    timerId = setTimeout(() => {
+      func(...args);
+    }, delay);
+  };
+}
```
ankushdharkar, line 48: Can this be a common util?

src/utils.js
```diff
@@ -0,0 +1,19 @@
...
+  const self_user = await res.json();
+  return self_user;
+}
+
+async function checkUserIsSuperUser() {
+  const self_user = await getSelfUser();
+  if (self_user?.roles['super_user']) {
+    return true;
+  }
+  return false;
```
ankushdharkar, line 18:
```suggestion
  return self_user?.roles['super_user'];
```

src/user-management/script.js
```diff
@@ -0,0 +1,127 @@
...
+    userListElement.appendChild(paraELe);
+    return;
+  }
+  const ulElement = document.createElement('ul');
+  users.forEach((userData) => {
+    const listElement = document.createElement('li');
+    const imgElement = document.createElement('img');
+    imgElement.src = userData.picture ? userData.picture : defaultAvatar;
+    imgElement.setAttribute('width', 70);
+    imgElement.setAttribute('height', 70);
```
ankushdharkar, line 97: Can these come from a class?

src/user-management/script.js
```diff
@@ -0,0 +1,127 @@
...
+    const pElement = document.createElement('p');
+    let name =
+      userData.first_name.slice(0, 15) + ' ' + userData.last_name.slice(0, 15);
```
ankushdharkar, line 100: Why is the name being sliced?

src/user-management/script.js
```diff
@@ -0,0 +1,127 @@
...
+    }
+    ulElement.appendChild(listElement);
+  });
+  addClass(loaderElement, 'element-display-remove');
+  userListElement.appendChild(ulElement);
+}
+
+function filterUsers(searchInput) {
+  usersListfilter = usersDataList.filter((user) => {
+    const name = user.first_name + (user.last_name ? user.last_name : '');
```
ankushdharkar, line 121:
```suggestion
    const name = `${user.first_name}  ${user.last_name ?? ''}`;
```

ankushdharkar (changes requested):

src/constants.js
poonammohata: ok, but once @Neha suggested that UI elements should also be in the constants file, in this issue it's mentioned(https://github.com/Real-Dev-Squad/website-dashboard/issues/203), though she suggested there should be different constants files for UI elements but for that we need to restructure the file structure(which I am working on for different issues), so, for now, I kept all the types constants in this file only.
Let me know if i understood something wrong

src/user-management/script.js
poonammohata: So the reason I am slicing here is, on users API on staging there was not so many users information, so I tested this feature with members API too, and there I found this user name
image
though this is not accurate data, in case if get such incorrect data so I sliced the name. 
But the slicing name is wrong, will remove it. I think we should have validation on the name when we are taking user information if in case we are not having it right now.

src/index.html
poonammohata: yes this should be an <a> tag, updated it

src/user-management/script.js
poonammohata: yes, added it in util

src/user-management/utils.js
```diff
@@ -0,0 +1,43 @@
+function addClass(element, className) {
+  element.classList.add(className);
+}
+
+function removeClass(element, className) {
+  element.classList.remove(className);
+}
+
```
ankushdharkar, line 8: Let's get rid of these one-liner abstractions. They add more overhead, than additional value

src/user-management/script.js
heyrandhir: Good observation. Instead of being declarative, we can let the last layer handle it. Read this https://developer.mozilla.org/en-US/docs/Web/CSS/text-overflow.

src/user-management/utils.js
```diff
@@ -0,0 +1,43 @@
...
+      headers,
+      credentials,
+      ...options,
+    });
+    return response;
+  } catch (err) {
+    console.error(MESSAGE_SOMETHING_WENT_WRONG, err);
+    return err;
+  }
+}
```
RitikJaiswal75, line 30: This can be a common util rather than keeping it in a separate user management util

src/user-management/utils.js
```diff
@@ -0,0 +1,43 @@
...
+  return (...args) => {
+    if (timerId) {
+      clearTimeout(timerId);
+    }
+
+    timerId = setTimeout(() => {
+      func(...args);
+    }, delay);
+  };
+}
```
RitikJaiswal75, line 43: Even this can go in common util

RitikJaiswal75 (changes requested):

src/user-management/utils.js
poonammohata: According to the current folder structure of this module, if I add it in common util, it will not work, I am working on fixing the folder structure in another issue, and once the folder structure will get fixed, I will add all functions which can be reused in common utils

src/user-management/utils.js
poonammohata: Right now, can't add it in common util, explained the reason in the above comment

heyrandhir (approved):

prakashchoudhary07:
We need to add the link to the repo 

Screenshot from 2022-11-15 05-18-01

src/index.html
```diff
@@ -27,6 +30,13 @@
       >
         Profile
       </button>
+      <a
+        id="user-management-link"
+        class="element-display-remove"
+        href="user-management/index.html"
```
ankushdharkar, line 36: Is this an absolute path or a relative path here @poonammohata?

src/script.js
```diff
@@ -0,0 +1,13 @@
+const userManagementLink = document.getElementById(USER_MANAGEMENT_LINK);
+export async function showUserManagementButton() {
+  try {
+    const isSuperUser = await checkUserIsSuperUser();
+    if (isSuperUser) {
+      userManagementLink.classList.remove('element-display-remove');
```
ankushdharkar, line 6: 1. Can `element-display-remove` be renamed something that talks about the actual behaviour of the class, and not _how_ it's doing this?
2. Maybe we should _add_ this class, when the user is a superuser, to provide additional features, so that most people never see a button flashing away?

src/style.css
```diff
@@ -79,6 +84,24 @@ button {
...
+#user-management-link {
+  color: black;
+  font-weight: 500;
+  font-size: larger;
+  background-color: var(--white-color);
+  border: 2px solid var(--black-color);
+  border-radius: 5px;
+  padding: 10px 50px;
+  cursor: pointer;
+  transition: all 0.5s ease;
```
ankushdharkar, line 96: Good work with the transition :)

src/user-management/script.js
```diff
@@ -0,0 +1,127 @@
...
+  listContainerElement.classList.remove('table-view');
+  listContainerElement.classList.add('tile-view');
+
+  listContainerElement.childNodes.forEach((listElement) => {
+    const imgElement = listElement.firstChild;
+    imgElement.classList.add('element-display-remove');
+    listElement.classList.add('tile-width');
+  });
+}
+
```
ankushdharkar, line 28: TODO: Talk about using config objects. - Ankush

src/user-management/utils.js
ankushdharkar: - [ ] ☝️ Please add ticket for the above mentioned refactor

ankushdharkar (changes requested):

ankushdharkar (approved):

ankushdharkar:
@poonammohata Merging this for now. Thank you.

Please take care of the mentioned unresolved comments later.
cc @heyrandhir

PR #245, opened 2022-11-12
added add tags feature

vinayak-trivedi:
- added a feature to add category and level to the user, which creates generic items tags for those tasks


src/task/index.html
```diff
@@ -106,7 +106,7 @@
             <option value="ASSIGNED">Assigned</option>
           </select>
         </div>
-        <div class="inputBox">
+        <div class="inputBox hidden">
```
vinayak-trivedi, line 109: hiding this until we have the api

heyrandhir (approved):
LGTM 🙂👍🏻

src/task/index.html
alishvagadiya: Add a comment in the file, instead of in PR,
No one goes back and check PR review comments.

ankushdharkar: @alishvagadiya I think the comment is meant for the reviewers of just this PR. It's probably there as an optional information.

ankushdharkar (approved):
Thank you for the changes.

PR #249, opened 2022-11-19
added admin-panel feature

vinayak-trivedi:
Note: Don't merge it for now, it won't create any problem but it would be better if it gets merged after depaiyan's generic model PR is merged
- added feature for admin panel
- current features:
   i the super user can create new skill tags
   ii. super user can create new levels


src/admin-panel/utils.js
```diff
@@ -0,0 +1,10 @@
+function createElement({ type, attributes = {}, innerText }) {
+  const element = document.createElement(type);
+  Object.keys(attributes).forEach((item) => {
+    element.setAttribute(item, attributes[item]);
+  });
+  element.textContent = innerText;
+  return element;
+}
+
+export { createElement };
```
ivinayakg, line 10: you have this same piece of code in the production, can't we make this into a generic util function and then import it her
Reference - https://github.com/Real-Dev-Squad/website-dashboard/blob/develop/src/taskEvents/utils.js#L3-L10

ivinayakg (changes requested):

src/admin-panel/utils.js
vinayak-trivedi: I think, this can be done in another PR, created an issue for it, we can assign it to any beginner https://github.com/Real-Dev-Squad/website-dashboard/issues/252

ivinayakg (dismissed):

src/admin-panel/index.js
```diff
@@ -0,0 +1,172 @@
...
+    type: 'button',
+    attributes: { class: 'submit-button' },
+    innerText: 'Create Level',
+  });
+  submit.addEventListener('click', createLevel);
+  levelCreationContainer.appendChild(title);
+  levelCreationContainer.appendChild(levelNameInput);
+  levelCreationContainer.appendChild(levelSelector);
+  levelCreationContainer.appendChild(submit);
+  return levelCreationContainer;
```
prakashchoudhary07, line 146: NIT: I think append can be used here `append(ele1, ele2, ele3..)`

src/admin-panel/index.js
```diff
@@ -0,0 +1,172 @@
...
+
+  for (let i = 0; i <= 5; i++) {
+    const option = createElement({
+      type: 'option',
+      attributes: { class: 'option', value: i },
+      innerText: i,
+    });
+    levelSelector.append(option);
+  }
+  return levelSelector;
```
prakashchoudhary07, line 114: NIT: appending elements can be done after loop is completed, and The times for loop runs should come from a constant

src/admin-panel/index.js
```diff
@@ -0,0 +1,172 @@
...
+    const body = {
+      name: inputValue,
+      type: 'SKILL',
+      reason: 'adding skills to users',
+    };
+
+    setLoadingState();
+    button.textContent = 'creating skill...';
+
+    const response = await fetch(`${BASE_URL}/tags`, {
```
prakashchoudhary07, line 80: NIT: URL should come from constant file

prakashchoudhary07 (approved):

prakashchoudhary07:
Shouldn't this be accessible only to Admin?

vinayak-trivedi:
> Shouldn't this be accessible only to Admin?

yes correct, done in the latest commit

prakashchoudhary07 (approved):

PR #250, opened 2022-11-20
added a feature to get suggested users based on skills

vinayak-trivedi:
NOTE: Don't merge this PR for now, there are other PRs in the backend which are need to be merged

- when the superuser is creating tasks they get an assignee field, where they can provide the assignee of the task, 
- This feature will allow them to get suggested users based on the skills that they have selected in the category field

### DIFF
--- a/src/task/script.js
+++ b/src/task/script.js
@@ -391,11 +391,11 @@ function debounce(func, delay) {
 async function fetchTags() {
   const response = await fetch(`${API_BASE_URL}/tags`);
   const data = await response.json();
-  const { allTags } = data;
+  const { tags } = data;
 
   const category = document.getElementById('category');
 
-  for (const tag of allTags) {
+  for (const tag of tags) {
     const option = document.createElement('option');
     option.textContent = tag.name;
     option.setAttribute('value', tag.id);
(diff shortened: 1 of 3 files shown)


src/task/script.js
```diff
@@ -461,12 +480,34 @@ function setAssignee(assignee) {
...
+    'suggested-users-container',
+  );
+
+  suggestedUsersContainer.innerHTML = '';
+  suggestedUsersDiv.classList.remove('hidden');
+
+  suggestedUsers.forEach((user) => {
+    const listItem = document.createElement('p');
+    listItem.classList.add('list-item');
+    listItem.style.cursor = 'pointer';
```
ivinayakg, line 501: if you are adding the class for the tag then, can't you just add the style directly through the CSS file using this particular classname, as putting styles through javascript can be a computation expensive task

ivinayakg (approved):

src/task/script.js
vinayak-trivedi: done

src/task/script.js
```diff
@@ -1,4 +1,25 @@
 const API_BASE_URL = window.API_BASE_URL;
+const suggestedUsers = [];
+const allUser = [];
+const category = document.getElementById('category');
+
+category.addEventListener('change', async () => {
+  try {
+    showSubmitLoader();
```
ivinayakg, line 8: are we disabling the select tag so that the user cannot change it while there is a request on the way.
or add a debounce function so that it will wait for lets say 0.5s before it makes the call and you can disable the select tag after the 0.5s

vinayak-trivedi: as per the discussion, I am already disabling that

src/task/script.js
```diff
@@ -370,32 +391,51 @@ function debounce(func, delay) {
...
+        levels[j + 1] = temp;
+        swapped = true;
+      }
+    }
+    if (!swapped) {
+      break;
+    }
+  }
+}
+
```
prakashchoudhary07, line 424: Can we use an inbuilt sort here?

prakashchoudhary07 (changes requested):

src/task/script.js
vinayak-trivedi: done

prakashchoudhary07 (approved):

PR #255, opened 2022-11-24
fix the bug remove src folder

### LINKED ISSUE #254: Remove `src` folder
Dashboard site was supposed to not use any build tools and be just raw version of HTML, CSS, JS to encourage devs to really become proficient in standard tech.

The `src` folder should hence be removed, as it's causing confusion in the URLs. Please make sure that all the pages of `src/` and at root level are correct.

vinit717:
Resolves #254 

remove the src folder and change all the base URL


index.html
```diff
@@ -16,21 +20,25 @@
...
       </button>
       <a
-        href="src/online-members/online-members.html"
-        class="online-members-link"
+        id="user-management-link"
+        class="element-display-remove"
+        href="user-management/index.html"
       >
+        User Management
+      </a>
```
vinayak-trivedi, line 40: why user management link is added in this PR?

vinit717: I did'nt add anything as there is 2 two index.html one in src folder and other one in main folder so as my task is to remove src folder to i'll remove 1 index.html but content is same in both except user management

prakashchoudhary07:
Please resolve the conflicts. 
And there are a few more files that are in src now after you have raised the PR, let's get those fixed and we should be good to go.

PR #259, opened 2022-12-02
Prevent fetching data on user input

### LINKED ISSUE #256: fetch the members data once and filter it on every  input
- currently, we are making API calls again and again on the input from the user, we should not be doing that
- we should just fetch the data once and filter it on every input

manish591:
fixes #256 

PROBLEM:
Please read about it in the above issue

SOLUTION:
On each user input, we are calling the `fetchMembers` function that will fetch all the users and filter the users based on the input query.

We have separated the filter member logic from the fetchMembers function and extracted it in a different function named `filterUser`.

We are calling fetchMembers only once and storing results in a variable called `membersData`. On each user input, we are calling the `filterMembers` function which will filter out users stored in the `membersData` variable.

### DIFF
--- a/src/task/script.js
+++ b/src/task/script.js
@@ -20,6 +20,7 @@ category.addEventListener('change', async () => {
     showSubmitLoader(false);
   }
 });
+let membersData = [];
 
 const getRemainingDays = (selectedDateInString) => {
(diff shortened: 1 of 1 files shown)


vinayak-trivedi (approved):
LGTM

PR #260, opened 2022-12-03
adds pagination and search functionality on user list

### LINKED ISSUE #248: Issues in User Management Listing Screen
- [ ] the user API is now paginated so by default only returns 100 users. Hence we are unable to see all the users currently. Hence we need to fetch the user based on the name being searched. we need to use `limit` and `skip` keywords to fetch the next 100 users in the user API.
- [ ] While searching the user the cards are not placed correctly. 
image

- [ ] By default users with archived roles shouldn't show up in the listing screen

poonammohata:
### Issue
Closes #248

### What are the changes?
- fixes user card placement
- showing only those user's on the listing screen whose archived role is false
- added pagination on the user listing screen: at a time only 100 users will be fetched if we want to see more users, we need to hit next button, on each click on next button, next 100 users will be fetched, and shown on the listing screen, to see the previous list of user we need to click the previous button, on the first page previous button is disabled and on the last page next button is disabled
- added search functionality of particular user by username

### Is Development Tested?
- yes

### Short video of how this will look on the website
https://user-images.githubusercontent.com/54801934/205439399-5ef0e490-71b7-48e4-a99d-1beb4d3bd69e.mp4

### DIFF
--- a/src/user-management/index.html
+++ b/src/user-management/index.html
@@ -40,7 +40,7 @@
       <div id="user-list">
         <div id="loader"></div>
       </div>
-      <div id="pagination" class="element-display-remove">
+      <div id="pagination" class="remove-element">
         <button class="pagination-btn" id="prevButton">&laquo; Previous</button>
         <button class="pagination-btn" id="nextButton">Next &raquo;</button>
       </div>
--- a/src/user-management/index.html
+++ b/src/user-management/index.html
@@ -17,6 +17,7 @@
     <section id="main-section">
       <div id="search">
         <input id="user-search" type="text" placeholder="Search Users" />
+        <p id="user-note">user can be searched by username only</p>
       </div>
       <div id="view-type-btn">
         <button id="tile-view-btn">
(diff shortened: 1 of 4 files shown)


Maheima:
Can we please change the colors of next and previous button so that it is more visible?

src/user-management/index.html
```diff
@@ -17,6 +17,7 @@ <h1>User Management</h1>
     <section id="main-section">
       <div id="search">
         <input id="user-search" type="text" placeholder="Search Users" />
+        <p id="user-note">user can be searched by username only</p>
```
heyrandhir, line 20:
```suggestion
        <p id="user-note">Currently the Users can be searched by their username only.</p>
```

heyrandhir (dismissed):

poonammohata:
> Can we please change the colors of next and previous button so that it is more visible?

changed the color, let me know if it needs any changes

In this case, the previous button is not active
image

In this case, both buttons are active
image

heyrandhir (approved):
LGTM

Maheima (dismissed):
LGTM!

Maheima (approved):
LGTM!

PR #263, opened 2022-12-03
fixed req body

vinayak-trivedi:
changed create level request body due to the new changes in backend

### DIFF
--- a/src/admin-panel/index.js
+++ b/src/admin-panel/index.js
@@ -4,7 +4,6 @@ const main = document.getElementById('main');
 const loading = document.getElementById('loading');
 const selectElement = document.getElementById('select-tags');
 const BASE_URL = 'https://api.realdevsquad.com';
-
 (async function setAuth() {
   try {
     const res = await fetch(`${BASE_URL}/users/self`, {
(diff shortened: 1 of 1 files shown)


src/admin-panel/index.js
```diff
@@ -84,7 +83,7 @@ async function createLevel() {
     alert(err);
   } finally {
     removeLoadingState();
-    button.textContent = 'Create Level';
+    document.querySelector('.submit-button').textContent = 'Create Level';
```
ankushdharkar, line 86: Why was this change needed?

src/admin-panel/index.js
```diff
@@ -118,7 +117,7 @@ async function createSkill() {
     alert(err);
   } finally {
     removeLoadingState();
-    button.textContent = 'Create Skill';
+    document.querySelector('.submit-button').textContent = 'Create Skill';
```
ankushdharkar, line 120: https://github.com/Real-Dev-Squad/website-dashboard/pull/263/files#r1038875892

ankushdharkar (changes requested):

src/admin-panel/index.js
vinayak-trivedi: because of my mistake in previous PR

ankushdharkar: On line 63, we already have a reference. `const button = document.querySelector('.submit-button');`, so why are we not using it here?

vinayak-trivedi: because it's scope is limited to the try block

ankushdharkar (approved):

PR #267, opened 2022-12-06
Feat/task user skill

Muralidhar22:
## change
- Refactoring level key from `level.name -> level.value`
- In Add skill modal, if a skill exists for a user it will not be shown in the options
image

### DIFF
--- a/src/taskEvents/script.js
+++ b/src/taskEvents/script.js
@@ -259,8 +259,9 @@ function openAddSkillModal(userId, skillsDiv, skills) {
       });
     }
     skillsDiv.append(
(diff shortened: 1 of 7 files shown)


src/taskEvents/script.js
```diff
@@ -347,7 +366,8 @@ async function render() {
       elementContainer.appendChild(element);
       container.appendChild(elementContainer);
     }
-  } catch (err) {
+  } catch (error) {
+    console.error(error);
```
vinayak-trivedi, line 370: here as well, can we remove the console statement

vinayak-trivedi (approved):
LGTM

Issue #289, opened 2023-01-23
Blank page in Incognito mode for unauthorized users


sahsisunny:
# Blank page in Incognito mode for unauthorized users

When accessing the dashboard site in a browser, unauthorized users are properly shown a "you are not authorized" message. However, when accessing the site in Incognito mode, the page appears blank and does not display any content.

### Steps to reproduce:
- Open a browser in Incognito mode
 - Navigate to the [Link](https://dashboard.realdevsquad.com/user-management/user-details/?username=ritik)

### Expected outcome:

- A "you are not authorized" message should be displayed

### Actual outcome:

- The page is blank and no content is displayed

**Note:** This issue needs to be fixed to ensure that unauthorized users receive the proper message when accessing the dashboard site in Incognito mode.

### Expected Output:

![Screenshot (42)](https://user-images.githubusercontent.com/70854507/213947954-6f3c1d49-35c7-4691-a6c5-835469571123.png)

sahsisunny:
fix is merged in #290 PR

PR #290, opened 2023-01-23
fix incognito unauthorized issue.

sahsisunny:
# Issue - #289 

## What is the change?
- Added a check to see if the user is authenticated or not. If the user is authenticated then only it will return self_user otherwise it will return nothing.
- one file is changed utils.js which is in the root directory of the project.
 
## Is Development Tested?
- [x] Yes

## After Change Screenshots

Screenshot (42)


heyrandhir (approved):

utils.js
```diff
@@ -7,7 +7,9 @@ async function getSelfUser() {
     },
   });
   const self_user = await res.json();
-  return self_user;
+  if (res.status === 200) {
+    return self_user;
+  }
```
RitikJaiswal75, line 12: What if i get 400 then nothing is sent back also my page goes blank like this

image

RitikJaiswal75 (changes requested):

utils.js
sahsisunny: done

RitikJaiswal75 (approved):
LGTM

utils.js
```diff
@@ -1,13 +1,17 @@
...
+      credentials: 'include',
+      headers: {
+        'Content-type': 'application/json',
+      },
+    });
+    const self_user = await res.json();
+    if (res.status === 200) {
+      return self_user;
+    }
+  } catch (err) {}
```
heyrandhir, line 14: we can console the error here instead of leaving the catch block empty.

heyrandhir (changes requested):

utils.js
sahsisunny: it is not necessary to show the error info to user.

heyrandhir (approved):

PR #293, opened 2023-01-24
Show 404 User Not Found

sahsisunny:
# Show 404 User Not Found

### What is the change?

- Two files changes in `/user/details` folder.
  - `index.js` : Added a function to show custom message for 404 error.
  - `style.css` : Added some css for custom message.
- Add one image file in `images` folder named `page-not-found.png`.

### Is Development Tested?

- [x] Yes

### After Changes

https://user-images.githubusercontent.com/70854507/216499679-451fb189-973d-48aa-89b5-913531143229.mp4


page-not-found/index.html
```diff
@@ -0,0 +1,41 @@
...
+  <head>
+    <meta charset="UTF-8" />
+    <meta http-equiv="X-UA-Compatible" content="IE=edge" />
+    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+    <link rel="stylesheet" href="style.css" />
+    <title>Page not found | Dashboard RDS</title>
+  </head>
+  <body>
+    <nav id="tasksNav">
+      <a id="createTask" class="nav-link"> Page Not Found </a>
```
heyrandhir, line 12: why are we adding an anchor link without href.

heyrandhir (changes requested):

heyrandhir (dismissed):

users/details/style.css
```diff
@@ -438,3 +438,6 @@ footer p a {
   display: flex;
   justify-content: center;
 }
+.not-found-img {
+  width: 55%;
```
shmbajaj, line 442: Also, add max-width style property for the image.

users/details/script.js
```diff
@@ -58,6 +61,25 @@ async function getUserData() {
   }
 }
 
+function dataNotFound(message) {
```
shmbajaj, line 64: If this function, is reusable outside the scope of `user/details`, if yes then please move to utils and change the method to `getDataNotFound` and then add to the container.
else if not then no change required.

shmbajaj (changes requested):

users/details/script.js
heyrandhir: I think @sahsisunny is only handling the Not Found for Users. Also a better name I could think of this function could be `generateNoUserFoundPage` or `generateNoDataFoundSection`

users/details/style.css
sahsisunny: added

users/details/script.js
sahsisunny: change function name to `generateNoDataFoundSection`

heyrandhir (approved):

shmbajaj (approved):
Next time, if possible create a constants.js file and store and get all raw strings from there.

Issue #326, opened 2023-02-27
User card shows pointer on hover


ankushdharkar:
# Cursor pointer on user card
## Overview

Anytime there is a clickable action on any element on the website, there should be a pointer cursor that appears when hovered. Let's start by adding it to our users management page.

## User stories

- As a viewer, there should be a pointer that appears on hover for each user that is clickable

## Requirements

- Cursor becomes a pointer when hovered on user card on user management page. You can [refer this MDN doc](https://developer.mozilla.org/en-US/docs/Web/CSS/cursor) for more details.
- TDD is encouraged to be followed
- Tests should be written

## Acceptance Criteria

- [ ] A pointer appears when hovering on the user card on user management page

PR #329, opened 2023-03-04
#326 Add cursor pointer in clickable elements

khushi818:
## Issues #326 
pointer added to 
- Buttons 
- input with type submit
- list elements for cards in user management

Instead of changing CSS in each style.css file, I created a global.css **as per the suggestions**

Note : Imports are replaced with links in html

### DIFF
--- a/featureFlag/style.css
+++ b/featureFlag/style.css
@@ -1,5 +1,3 @@
-@import url('../global.css');
-
 * {
   list-style: none;
   text-decoration: none;
--- a/featureFlag/style.css
+++ b/featureFlag/style.css
@@ -1,3 +1,5 @@
+@import url('../global.css');
+
 * {
   list-style: none;
   text-decoration: none;
@@ -17,10 +19,6 @@
 body {
   display: flex;
   align-items: center;
-}
-
-button {
-  cursor: pointer;
 }
 
 .wrapper {
--- a/featureFlag/style.css
+++ b/featureFlag/style.css
@@ -19,6 +19,10 @@ body {
   align-items: center;
 }
 
+button {
+  cursor: pointer;
+}
+
 .wrapper {
   width: clamp(300px, 90%, 800px);
 }
@@ -103,6 +107,7 @@ body {
   height: 20px;
   width: 20px;
 }
+
 .checkbox__tick-box--reversed {
   margin-left: 20px;
   margin-right: 0;
--- a/goal/style.css
+++ b/goal/style.css
@@ -25,6 +25,11 @@ body {
   user-select: none;
 }
 
+button,
+input[type='submit'] {
+  cursor: pointer;
+}
+
 .container {
   width: 100%;
   color: var(--color-gray);
--- a/goal/style.css
+++ b/goal/style.css
@@ -1,3 +1,5 @@
+@import url('../global.css');
+
 :root {
   --color-white: #fff;
   --color-black: #000;
@@ -23,11 +25,6 @@ body {
   flex-direction: column;
   height: 100vh;
   user-select: none;
-}
-
-button,
-input[type='submit'] {
-  cursor: pointer;
 }
 
 .container {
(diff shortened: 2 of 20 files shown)


featureFlag/style.css
```diff
@@ -19,6 +19,10 @@ body {
   align-items: center;
 }
 
+button {
```
sahsisunny, line 22: Can we use `class` name instead of `tag`?

khushi818: Yes, we can but buttons are defined at the top level and could be overridden very easily. If new buttons or input with submit type is added, they will automatically have a pointer cursor and could be changed by defining using their class or id.

TanishqSingla (approved):

featureFlag/style.css
sahsisunny: If you are setting a universal properties for all buttons and elements then you can create a global CSS file also

khushi818: got it thanks 👍

khushi818: @sahsisunny I did the changes you asked

sahsisunny (dismissed):

featureFlag/style.css
```diff
@@ -1,3 +1,5 @@
+@import url('../global.css');
```
TanishqSingla, line 1: Import statement might not be ideal here, can this be changed?
refer [this](https://www.debugbear.com/blog/avoid-css-import#:~:text=The%20CSS%20%40import%20rule%20can,to%20take%20longer%20to%20render.)

goal/style.css
```diff
@@ -1,3 +1,5 @@
+@import url('../global.css');
```
TanishqSingla, line 1: same as above

online-members/online-members.css
```diff
@@ -1,3 +1,5 @@
+@import url('../global.css');
```
TanishqSingla, line 1: same as above

profile/style.css
```diff
@@ -1,3 +1,5 @@
+@import url('../global.css');
```
TanishqSingla, line 1: same as above

style.css
```diff
@@ -1,8 +1,11 @@
+@import url('../global.css');
```
TanishqSingla, line 1: same as above

task/style.css
```diff
@@ -1,3 +1,5 @@
+@import url('../global.css');
```
TanishqSingla, line 1: same as above

taskEvents/style.css
```diff
@@ -1,3 +1,5 @@
+@import url('../global.css');
```
TanishqSingla, line 1: same as above

users/details/style.css
```diff
@@ -1,3 +1,5 @@
+@import url('../../global.css');
```
TanishqSingla, line 1: same as above

users/style.css
```diff
@@ -1,3 +1,5 @@
+@import url('../global.css');
```
TanishqSingla, line 1: same as above

wallet/style.css
```diff
@@ -1,3 +1,5 @@
+@import url('../global.css');
```
TanishqSingla, line 1: same as above

TanishqSingla (changes requested):

TanishqSingla (approved):

sahsisunny (approved):
Looks good to me

kotesh-arya (approved):
Looks Good To Me 👍🏻

featureFlag/style.css
kotesh-arya: > Import statement might not be ideal here, can this be changed? refer [this](https://www.debugbear.com/blog/avoid-css-import#:~:text=The%20CSS%20%40import%20rule%20can,to%20take%20longer%20to%20render.)

Great observation and well-suggested guideline @TanishqSingla  🙌🏻

Issue #325, opened 2023-02-27
Footer to link to our repo


ankushdharkar:
# Add footer
## Overview

Every website should have a footer message that links the viewer to our GitHub repo and help them discover the hard work that went behind creating the apps. 

## User stories

- As a viewer, there should be a footer on all pages of Dashboard site that looks similar to our other websites:

<img width="584" alt="image" src="https://user-images.githubusercontent.com/1935403/221709492-215fd81f-a648-4500-aba0-c9c25086edee.png">

## Requirements

- Footer should link back to dashboards site repo.
- TDD is encouraged to be followed
- Tests should be written

## Acceptance Criteria

- [ ] A footer appears on every dashboard site webpage which contains a link to our repo for dashboards app.

PR #330, opened 2023-03-05
Test/footer

khushi818:
## Issue #325


sahsisunny (approved):

footerTest.js
```diff
@@ -0,0 +1,70 @@
...
+        // will return undefined if the element is not found
+      });
+      console.log(`✅ footer text:${footer_text}  index:${index}:${url}`);
+
+      if (footer_text) {
+        // links inside footer are working or not
+        await Promise.all([
+          page.click('.info-repo a', { delay: 2000 }),
+          page.waitForNavigation(),
+          setTimeout(() => {
```
ankushdharkar, line 50: This is a slightly hacky solution.

Let's say our test runs in 1s. Now, we will have to wait extra 9s for the test to exit. If we are running 100+ tests, then we would have an additional 15min delay on our CI for such wait times.

Additionally, if our test was going to run in 11s, your setTimout closes the browser in 10s and reports the test as failed.(False Negative -- Though this is a lesser problem, since we have test timeouts setup anyway)

ankushdharkar: Let's refactor some of it using the working setup in [this branch](https://github.com/Real-Dev-Squad/website-tests/tree/tmp-ankush-single-test-working)

footerTest.js
```diff
@@ -0,0 +1,70 @@
...
+    } else {
+      console.log(`❌no footer at ${url}`);
+      browser.disconnect();
+    }
+  });
+};
+
+async function runTest() {
+  for (let url of urls) {
+    console.log(urls.indexOf(url));
```
ankushdharkar, line 65: Let's avoid console logs?

footerTest.js
```diff
@@ -0,0 +1,70 @@
...
+      }
+    } else {
+      console.log(`❌no footer at ${url}`);
+      browser.disconnect();
+    }
+  });
+};
+
+async function runTest() {
+  for (let url of urls) {
```
ankushdharkar, line 64: If I am reading this correctly, you are spinning up a new browser to test every url. Right?

While this works, there's a faster way using the "incognito" mode of browser. The browser is already loaded and ready to go, and we create a new isolated context per page, which loads faster (We should do some benchmarking around this!). See some examples of this [here](https://github.com/Real-Dev-Squad/website-tests/blob/ada05cc3d0a9f14923359d2c828e8b48988e4e3c/acceptance/new-user/www-loads.test.ts#L11).

ankushdharkar (changes requested):

footerTest.js
khushi818: I did that but I was facing a problem closing the browser. will try again.

footerTest.js
khushi818: Okay got it

footerTest.js
khushi818: Yup will go through it

khushi818: @ankushdharkar can I use headless chrome? All the problems resolved you mentioned using it.

style.css
```diff
@@ -42,6 +42,7 @@ body {
 
 button {
   width: 250px;
+  cursor: pointer;
```
TanishqSingla, line 45: we already have `cursor: pointer` in globals.css why do we need it here?

online-members/online-members.css
```diff
@@ -112,3 +112,14 @@
   display: flex;
   align-items: center;
 }
+
+footer {
+  display: block;
+  width: 100%;
```
TanishqSingla, line 118: I don't think we need `width: 100%` since we already have display block

online-members/online-members.css
```diff
@@ -112,3 +112,14 @@
...
+  padding: 8px;
+}
+
+footer .info-repo {
+  font-weight: 100;
+  text-align: center;
+}
```
TanishqSingla, line 125: Is there a reason we're styling the footer here differently than css of other pages.

users/style.css
```diff
@@ -278,6 +278,16 @@ li p {
...
 }
 
+.info-repo {
+  font-weight: 100;
+  padding: auto;
+  text-align: center;
+}
+
+footer {
+  margin-top: auto;
```
TanishqSingla, line 288: does the auto have any effect here?

users/style.css
```diff
@@ -278,6 +278,16 @@ li p {
   display: none;
 }
 
+.info-repo {
+  font-weight: 100;
+  padding: auto;
```
TanishqSingla, line 283: does the auto have any effect here?

featureFlag/style.css
```diff
@@ -123,6 +123,16 @@ body {
...
 }
 
+.info-repo {
+  font-weight: 100;
+  padding: 8px;
+  text-align: center;
+}
+
+footer {
+  margin-top: auto;
```
TanishqSingla, line 133: what effect does `auto` add here? do we need this property?

TanishqSingla (changes requested):

style.css
khushi818: This was not done by me. This change was already there.

online-members/online-members.css
khushi818: I just copy-paste the CSS for now from other sites. As the main task was testing.

footerTest.js
```diff
@@ -0,0 +1,68 @@
...
+    const page = await context.newPage();
+    await page.goto(url, { waitUntil: ['load', 'networkidle2'] });
+    await page.content();
+    page.setDefaultNavigationTimeout(0);
+    const footer = await page.$('footer');
+    if (footer) {
+      // is the text same in footer?
+      const footer_text = await page.evaluate(() => {
+        const element = document.querySelector('.info-repo');
+        console.log(element);
```
prakashchoudhary07, line 35: console.log?

footerTest.js
```diff
@@ -0,0 +1,68 @@
...
+  });
+};
+
+async function runTest() {
+  for (let url of urls) {
+    await test_footer(url, urls.indexOf(url));
+
+    if (urls.indexOf(url) === urls.length - 1) {
+      process.exit(0);
+    }
```
prakashchoudhary07, line 64: Are we launching multiple browsers for checking all URL's
Is that for running it parallel?

featureFlag/index.html
```diff
@@ -108,5 +108,16 @@
...
+      <p class="info-repo">
+        The contents of this website are deployed from this
+        <a
+          href="https://github.com/Real-Dev-Squad/website-dashboard"
+          target="_blank"
+          rel="noopener noreferrer"
+          >open sourced repo</a
+        >
+      </p>
+    </footer>
```
prakashchoudhary07, line 121: Let us extract the footer code in one place and use it from there. Let's create a separate issue and track this

prakashchoudhary07 (changes requested):

footerTest.js
khushi818: Yes, Checking all URLs. It is synchronous and it is headless, you can check the index in the above image I have shared.

featureFlag/index.html
khushi818: sure we can. It has vanilla code so there is duplicate code everywhere.

footerTest.js
khushi818: thanks, removed now

prakashchoudhary07 (approved):

Issue #333, opened 2023-03-18
Refactor | Replace redundant API URLs using constants. 


RitikJaiswal75:
# Issue
Our dashboard site currently has a base constants file with `API_BASE_URL`, which was supposed to be imported into other files wherever it was required.

But I ran a global search for the URL and found every folder has a separate constant.js and the `API_BASE_URL` is being rewritten rather than using the pre-created constant.

### Image of the issue:
![image](https://user-images.githubusercontent.com/57758447/226118691-79aedc32-35d0-4623-8491-fc334e30dee3.png)

Srinja333:
Can I work on this?

mukadas026:
@RitikJaiswal75 Please assign this to me

imsidkg:
Can I work on this isssue?

S1ddG:
Hi @RitikJaiswal75 Would love to fix this one!

musukeshu:
hi @RitikJaiswal75 . I would like to fix this one

sahsisunny:
@FMK2312 why you open this issue ?

PR #334, opened 2023-03-20
[Issue 333] Refactored / replaces redundant API URLs using global constants file

musukeshu:
Replaced BASE_API_URL wherever used locally to the one in the constants.js file.
Tested all the files wherever constants (BASE_API_URL) was refactored with live server.

https://github.com/Real-Dev-Squad/website-dashboard/issues/333


sahsisunny (commented):
- Link the Issues in the PR
- Three more API show in the Files can we export and import the `API_BASE_URL` and use it like `{API_BASE_URL}/members` ?
image

musukeshu:
Instead of using import and export we can directly use it by adding the global constants.js as a  script in the html file.
And updated the online-members/constants.js file.

I don't think in loadEnv.js we have to change as it is setting the URL for the window.API_BASE_URL, where its better not to refer it from global constants.

sahsisunny (approved):

samyakshah3008 (approved):
Great work, looks good to merge 👌

PR #353, opened 2023-04-28
filter api fix and footer alignment

heyrandhir:
- [x] Please merge this after [PR](https://github.com/Real-Dev-Squad/website-backend/pull/1065) and dev to main sync of backend. 
- [x] Fix the filter showing wrong results (the endpoint was not correct it was still hitting the old api not the endpoint which was created for search filters)
- [x] Position the footer at bottom when there isn't enough content. 
- [x] Position the previous and next buttons at bottom when there isn't enough content. 

*Earlier*
image

*Now*
image

### DIFF
--- a/users/script.js
+++ b/users/script.js
@@ -271,7 +271,10 @@ async function getParticularUserData(
 
 function showUserList(users) {
   const ulElement = document.createElement('ul');
-
+  if (!users.length) {
+    userListElement.innerHTML = 'No Users Found';
+    return;
+  }
   users.forEach((userData) => {
     const listElement = document.createElement('li');
     const imgElement = document.createElement('img');
--- a/users/script.js
+++ b/users/script.js
@@ -278,7 +278,9 @@ function showUserList(users) {
     imgElement.src = userData.picture?.url ?? DEFAULT_AVATAR;
     imgElement.classList.add('user-img-dimension');
     const pElement = document.createElement('p');
-    const node = document.createTextNode(`${userData.full_name}`);
+    const node = document.createTextNode(
+      `${userData.first_name} ${userData.last_name}`,
+    );
     pElement.appendChild(node);
     listElement.appendChild(imgElement);
     listElement.appendChild(pElement);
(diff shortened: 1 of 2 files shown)


users/script.js
```diff
@@ -278,7 +278,9 @@ function showUserList(users) {
     imgElement.src = userData.picture?.url ?? DEFAULT_AVATAR;
     imgElement.classList.add('user-img-dimension');
     const pElement = document.createElement('p');
-    const node = document.createTextNode(`${userData.full_name}`);
+    const node = document.createTextNode(
+      `${userData.first_name} ${userData.last_name}`,
```
vikhyat187, line 282: @heyrandhir is there any reason we changed `full name` to `first + last name`?

users/style.css
```diff
@@ -27,6 +27,27 @@
   color: var(--white-color);
 }
 
+html,
```
vikhyat187, line 30: @heyrandhir if you can share the screenshots before and after it would be really helpful.

users/script.js
heyrandhir: yes earlier in filter we only had filter for status i.e OOO , idle. that status api was doing the same thing and returning from backend. the current api is returning the same as it is stored in the users collection. i.e first name and last name so I had to do it here.

users/style.css
heyrandhir: sure I will add it

heyrandhir: Done @vikhyat187

vikhyat187 (approved):

vinit717 (approved):
LGTM

vikhyat187 (approved):

RitikJaiswal75:
The footer is ok to bring down. But, getting the next button completely at Botton doesn't feel right

users/script.js
```diff
@@ -271,14 +271,19 @@ async function getParticularUserData(
 
 function showUserList(users) {
   const ulElement = document.createElement('ul');
-
+  if (!users.length) {
+    userListElement.innerHTML = 'No Users Found';
+    return;
+  }
```
RitikJaiswal75, line 277: currently, this check will not work as our user API is also returning archived users. but a good thought to keep it.

RitikJaiswal75 (approved):

sahsisunny (approved):
Ship it!😜

users/script.js
heyrandhir: yeah this is. for the case when the applied filter doesn't gives any user.

Issue #357, opened 2023-05-07
inDiscord Feature - Frontend


bhtibrewal:
### User- inDiscord Feature  [#139](https://github.com/Real-Dev-Squad/todo-action-items/issues/139)
- [x] User inDiscord Feature - UI #347 
- [x] change URL to /users/discord and put the feature behind feature flag. #360 
- [ ] Integrate APIs and remove the dummy data.
- [ ] Write tests

PR #363, opened 2023-05-20
Integrate APIs and remove the dummy data.

bhtibrewal:
issue: #357


__tests__/users/App.test.js
```diff
@@ -53,15 +64,25 @@ describe('App Component', () => {
...
     expect(tabsSection).toBeDefined();
+    const tabs = await tabsSection.$$('.tab');
+    expect(tabs.length).toEqual(2);
 
     expect(usersSection).toBeDefined();
 
     expect(userDetailsSection).toBeDefined();
   });
+
+  it('should update the URL query string and re-render the app', async () => {
```
bhtibrewal, line 80: need to write more tests will be addressed in #357

users/discord/index.js
```diff
@@ -1,7 +1,7 @@
 import { App } from './App.js';
-const urlParams = new URLSearchParams(window.location.search);
+// const urlParams = new URLSearchParams(window.location.search);
 
-if (!urlParams.get('dev')) history.back();
+// if (!urlParams.get('dev')) history.back();
```
bhtibrewal, line 4: had to remove the `dev=true` flag to show the selected tab in URL query

__tests__/users/App.test.js
```diff
@@ -27,24 +28,34 @@ describe('App Component', () => {
 
     page.on('request', (interceptedRequest) => {
       const url = interceptedRequest.url();
-      if (url === `${API_BASE_URL}/users/inDiscord`) {
+      if (url === `${API_BASE_URL}/users/search/?role=in_discord`) {
```
iamitprakash, line 31: this line is checking if the URL matches the pattern ${API_BASE_URL}/users/search/?role=in_discord. However, this pattern will only match URLs that contain the exact string in_discord in the query string. This means that URLs that contain other values for the role parameter, such as moderator or admin, will not be matched.

iamitprakash, line 31: this line is checking if the URL matches the pattern ${API_BASE_URL}/users/search/?role=in_discord. However, this pattern will only match URLs that contain the exact string in_discord in the query string. This means that URLs that contain other values for the role parameter, such as moderator or admin, will not be matched.

iamitprakash (changes requested):

__tests__/users/App.test.js
bhtibrewal: Yes Tejas it will only match this specific API and that is the requirement as we are only calling only `${API_BASE_URL}/users/search/?role=in_discord` from frontend in according to the feature, tests for other roles and API functionality I have written in the backend.

iamitprakash: cool then, i am resolving this and leaving for another reviewer

iamitprakash (approved):
Here you go.......!

vvaibhavdesai (approved):

PR #378, opened 2023-06-15
Fix CSS of extension-requests page

Aryex82:
### Description:
Changed the CSS to fix the cards displayed extension requests page.

### Dev Tested:
- [x] Yes

### Images/video of the change:

before: 

image

after: 

https://github.com/Real-Dev-Squad/website-dashboard/assets/98796547/347df4b8-eb85-460f-a459-e161bce60828

## Follow-up Issues (if any)

### DIFF
--- a/extension-requests/style.css
+++ b/extension-requests/style.css
@@ -40,7 +40,7 @@ body {
 .container {
   width: 100%;
   text-align: center;
-  padding: 10px;
+  padding: 40px;
 }
 
 .header {
@@ -52,6 +52,7 @@ body {
 .extension-requests-modal-parent {
   display: none;
 }
+
 .extension-requests-modal-parent.visible {
   z-index: 1;
   position: fixed;
@@ -137,10 +138,12 @@ body {
   border: none;
   border-bottom: 1px solid var(--gray);
 }
+
 .extension-requests-form input:focus {
   outline: none;
   border-bottom: 1px solid var(--gray);
 }
+
 .extension-requests-form button {
   background-color: var(--button-proceed);
   border: none;
@@ -168,10 +171,12 @@ body {
   border: none;
   border-bottom: 1px solid var(--gray);
 }
+
 .extension-requests-status-form input:focus {
   outline: none;
   border-bottom: 1px solid var(--gray);
 }
+
 .extension-requests-status-form button {
   background-color: var(--button-proceed);
   border: none;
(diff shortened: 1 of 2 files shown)


sahsisunny:
image
Can you style the button in one ROW with equal width?

sahsisunny:
And also share videos that show responsiveness.

extension-requests/style.css
sahsisunny, line 1: Please use REM instead for PX if the value is greater than 8px.

Aryex82:
> And also share videos that show responsiveness.

added

Aryex82:
> image
> Can you style the button in one ROW with equal width?

done

extension-requests/style.css
```diff
@@ -40,7 +40,7 @@ body {
 .container {
   width: 100%;
   text-align: center;
-  padding: 10px;
+  padding: 40px;
 }
```
Pratiyushkumar, line 44: please use rem instead of px, use px only where the size is fixed for all screen-sizes.

extension-requests/style.css
```diff
@@ -187,42 +192,53 @@ body {
...
 
 .extension-request {
+  display: flex;
+  flex-direction: column;
+  justify-content: space-between;
+  overflow: auto;
   text-align: left;
   border: 2px solid var(--black-transparent);
   padding: 1.25rem;
```
Pratiyushkumar, line 209: also try using even measurements which helps in proper debugging and makes code look good

Pratiyushkumar (changes requested):
as commented by Sunny https://github.com/Real-Dev-Squad/website-dashboard/pull/378#issuecomment-1594187682
it says both the button should be in one row and with equal width and when I see the SS  the issue looks similar as before also share a clear video of the feature because video is blur

extension-requests/style.css
Aryex82: done

Aryex82:
> as commented by Sunny [#378 (comment)](https://github.com/Real-Dev-Squad/website-dashboard/pull/378#issuecomment-1594187682)
> it says both the button should be in one row and with equal width and when I see the SS the issue looks similar as before also share a clear video of the feature because video is blur

Updated the video. Buttons when the were of equal length looked a bit odd to me, anyways I've changed it to fixed length now.

extension-requests/style.css
Aryex82: Im not sure what you are referring to here. Can you please elaborate more.

sahsisunny (approved):
LGTM🎉🎉

Aryex82:
> as commented by Sunny [#378 (comment)](https://github.com/Real-Dev-Squad/website-dashboard/pull/378#issuecomment-1594187682)
> it says both the button should be in one row and with equal width and when I see the SS the issue looks similar as before also share a clear video of the feature because video is blur

done

Pratiyushkumar (approved):
LGTM

PR #437, opened 2023-07-31
Feat: Update query params on the browser url

j24m:
## Description :
- Implemented a mechanism to update the URL parameters when filters are applied, reflecting the filtered state on browser URL.
- Persisting user data & filters based on query parameters.
- Handling case sensitivity for query parameter values.
- Persisting the filters and user data when the url(along with the query params) is shared with other people or opened on a new tab.
- Clearing the query params when clicked on clear filters button.
## Video with changes :

https://github.com/Real-Dev-Squad/website-dashboard/assets/53934353/39f20b33-90c5-4252-908f-a760e60b9eed

### DIFF
--- a/users/script.js
+++ b/users/script.js
@@ -543,21 +543,6 @@ function parseQueryParams() {
   return queryObject;
 }
 
-// function selectFiltersBasedOnQueryParams(filters) {
-//   const parsedQuery = parseQueryParams();
-//   const checkBoxArray = Array.from(filters);
-//   checkBoxArray.map((filter) => {
-//     for (let key in parsedQuery) {
-//       const checkedValues = parsedQuery[key];
-//       checkedValues.map((value) => {
-//         if (filter.id === value) {
-//           filter.checked = true;
-//         }
-//       });
-//     }
-//   });
-// }
-
 function selectFiltersBasedOnQueryParams(filters) {
   const parsedQuery = parseQueryParams();
 
--- a/users/script.js
+++ b/users/script.js
@@ -515,10 +515,19 @@ function manipulateQueryParamsToURL(constructedQueryParam) {
   }
 }
 
+/**
+ * Parses the query parameters from the current URL and organizes them into an object.
+ *
+ * @function
+ * @returns {Object.<string, string[]>} An object containing query parameter keys as properties
+ * and arrays of corresponding values. Values are either in uppercase (for 'state') or as provided in the URL.
+ */
 function parseQueryParams() {
-  const urlString = window.location.href;
-  const urlObjInstance = new URL(urlString);
+  // Create a new URL object instance based on the current URL of the web page.
+  const urlObjInstance = new URL(window.location.href);
+  // Get an iterator for the query parameters.
   const queryParamsObj = urlObjInstance.searchParams.entries();
+  // Create an empty object to store the parsed query parameters.
   const queryObject = {};
   for (const [key, value] of queryParamsObj) {
     if (!queryObject[key]) {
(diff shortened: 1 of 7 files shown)


users/script.js
```diff
@@ -474,18 +485,76 @@ function getCheckedValues(groupName) {
...
   });
-
   checkedValuesAvailability.forEach((availability) => {
     params.append('state', availability);
   });
-
   return `?${params.toString()}`;
 }
 
+function updateQueryParamsToURL(constructedQueryParam) {
```
yesyash, line 497: - should we not name this `setQueryParams` / `addQueryParams` as we are setting the query params in the URL?

users/script.js
```diff
@@ -474,18 +485,76 @@ function getCheckedValues(groupName) {
...
+  const currentURLInstance = new URL(window.location.href);
+  currentURLInstance.search = '';
+  const currentURL = currentURLInstance.href;
+  const newUrlWithoutQueryParams = `${currentURL}`;
+  window.history.pushState(
+    { path: newUrlWithoutQueryParams },
+    '',
+    newUrlWithoutQueryParams,
+  );
+}
```
yesyash, line 519: - most of the logic inside both the functions look same can we abstract this out into one function and just pass a URL to that ?

iamitprakash (approved):

users/script.js
```diff
@@ -474,18 +485,92 @@ function getCheckedValues(groupName) {
...
+    if (!queryObject[key]) {
+      queryObject[key] = [];
+    }
+    if (key === 'state') {
+      queryObject[key].push(value.toLocaleUpperCase());
+    } else {
+      queryObject[key].push(value);
+    }
+  }
+  return queryObject;
```
yesyash, line 533: - Why are we doing this? What will the `queryObject` look like at the end? 
Can you please add a JSDoc comment stating the return type.

users/script.js
```diff
@@ -474,18 +485,92 @@ function getCheckedValues(groupName) {
...
+}
+
+function selectFiltersBasedOnQueryParams(checkboxes) {
```
yesyash, line 536: - Should the param here be named `filters` instead of `checkboxes`?

users/script.js
```diff
@@ -474,18 +485,92 @@ function getCheckedValues(groupName) {
...
+  checkBoxArray.map((checkbox) => {
+    for (let key in parsedQuery) {
+      const checkedValues = parsedQuery[key];
+      checkedValues.map((value) => {
+        if (checkbox.id === value) {
+          checkbox.checked = true;
+        }
+      });
+    }
+  });
```
yesyash, line 548: - Array.from give us a shallow copy of the original array, changing the nested values might cause a change to the original array can we cross-check this one? 
- Can we think of any other way to set the `checked` status of the filters for each query param?
- If you just want to iterate over an array I think using a `forEach` loop would be more accurate, also for iterating over an object where you only need the keys for it you can make use of Object.entries.

- Please add a JSDoc comment stating the type of `checkboxes` param the function is taking.

users/script.js
j24m:
```
/**
 * Parses the query parameters from the current URL and organizes them into an object.
 *
 * @function
 * @returns {Object.<string, string[]>} An object containing query parameter keys as properties
 * and arrays of corresponding values. Values are either in uppercase (for 'state') or as provided in the URL.
 */
```

users/script.js
j24m: - Optimized this function in the recent commit.
### JsDoc comment stating the type of checkbox(changed to filters) :
`@param` {NodeList} filters - A NodeList of filter elements with 'id' and 'checked' properties.

users/script.js
yesyash, line 1: - Please remove `console.log` from the file. If needed please use `console.debug`.

yesyash (approved):

Issue #430, opened 2023-07-27
Revamp UI & UX of Extensions Requests page


Aryex82:
Tasks: 
- [ ] Add functionality to filter ER by status (PENDING, ACCEPTED, DENIED). Page should display PENDING ER by default. 
- [ ] Add functionality to sort ER by created date
- [ ] Improve UI & UX of ER Cards 
- [ ] Add Error Handling

PR #459, opened 2023-08-17
Add functionality to update the Extension card.

Aryex82:
Issue : #430  

### Why is the change?
Previously user had to click on edit option which would open a modal and then update the required fields which felt unnecessary.

### What is the change?
This PR adds a button when clicked turns the required fields to editable. User can then choose to save the updated values or cancel it.

### Refactors: 
- Added new css classes to old elements.

### Note to Reviewers:
- The feature is under a feature flag. To access, please use `?dev=true` query param.
-  
### Test stats

Screenshot 2023-08-17 at 9 39 23 PM

### \*Dev Tested?

- [x] Yes
- [ ] No

### After Change video: 

https://github.com/Real-Dev-Squad/website-dashboard/assets/98796547/8716249f-53a0-4791-9e63-062eb2969697


yesyash (approved):

extension-requests/style.css
```diff
@@ -723,6 +787,17 @@ body {
...
 
+  .update-wrapper {
+    flex-direction: column;
+  }
+  .update-button {
+    font-size: 10px;
+  }
+
+  .cancel-button {
+    font-size: 10px;
```
isVivek99, line 798: aren't we using BEM here?

Aryex82: No, not followed in this repo

isVivek99 (approved):

Issue #511, opened 2023-09-12
Onboarding> 31 days failing from backend API


iamitprakash:
As a superuser i want to see user with multiple status.

In below edge case i have selected both Active and Onboardingstatus>31 error has been observed below

{
    "statusCode": 400,
    "error": "Bad Request",
    "message": "\"state[1]\" must be one of [OOO, ONBOARDING, IDLE, ACTIVE]"
}
![Screenshot 2023-09-12 at 1 16 33 PM](https://github.com/Real-Dev-Squad/website-dashboard/assets/34869115/3e02b602-4712-42a0-9618-0f840b11606d)

RitikJaiswal75:
Hey @iamitprakash can you confirm if the api used here is /users/search

Shenoy07:
Assigned to me.

PR #525, opened 2023-09-24
Filtering Fixed For Onboarding > 31 feature

Shenoy07:
Hello,

This is a fix for issue #511, the issue ticket says that it is a backend issue but after carefully reviewing the code with the creator, the problem was found in the front end.

While writing the code we did not consider the case where we select multiple checkboxes. The code only covered the part where the checkbox for "onboarding > 31 days" is checked. Look at the below images for a better understanding.

**Image 1:** Working of the current page when 'onboarding>31' is **NOT** selected.
image

**Image 2:** Working of the page when 'onboarding>31' **is** selected. It sends a wrong API call.
image

Only adding the other cases in the code was not enough, as after refreshing the page we ran into one more issue.

I had to write a function to change the query parameters if the onboarding>31 checkbox is checked. The function returns the 
updated query parameters so that the right API call is made.

Post making the required changes, we can see the right API calls being made.
image

**TESTS:**
Tests are passing as we can see in the video below:

https://github.com/Real-Dev-Squad/website-dashboard/assets/31709147/1a50b1fa-6e38-48f2-acd6-5d306cce572d

### DIFF
--- a/users/script.js
+++ b/users/script.js
@@ -400,11 +400,10 @@ function populateAvailability() {
     { name: 'Ooo (Out of Office)', id: 'OOO' },
     { name: 'Idle', id: 'IDLE' },
     { name: 'Onboarding', id: 'ONBOARDING' },
-    { name: 'Onboarding > 31d', id: 'ONBOARDING31DAYS' },
   ];
 
-  if (params.get('dev') != 'true') {
-    availabilityArr.pop();
+  if (params.get('dev') === 'true') {
+    availabilityArr.push({ name: 'Onboarding > 31d', id: 'ONBOARDING31DAYS' });
   }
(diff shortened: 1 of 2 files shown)


users/script.js
```diff
@@ -400,11 +400,10 @@ function populateAvailability() {
     { name: 'Ooo (Out of Office)', id: 'OOO' },
     { name: 'Idle', id: 'IDLE' },
     { name: 'Onboarding', id: 'ONBOARDING' },
-    { name: 'Onboarding > 31d', id: 'ONBOARDING31DAYS' },
   ];
```
Shenoy07, line 404: Adding it when dev === 'true' rather than removing it.

iamitprakash (approved):

Issue #531, opened 2023-09-27
Enhance Mobile Navbar Behavior and Styling


sahsisunny:
**Description**:
The current mobile navigation bar on dashboard.realdevsquad.com has two key issues that need to be addressed:

1. The hamburger menu does not close when clicking anywhere on the screen, which results in a poor user experience.
2. The hamburger icon and the background color of the navigation bar are the same, causing a lack of visual distinction.

**Steps to Reproduce**:
1. Visit dashboard.realdevsquad.com on a mobile device.
2. Click the hamburger icon to open the navigation menu.
3. Observe that clicking anywhere on the screen does not close the menu.
4. Notice that the hamburger icon and the background color of the navigation bar are the same, making it hard to distinguish the menu icon.

**Expected Behavior**:
1. When the user clicks anywhere on the screen, outside of the navigation menu, the hamburger menu should close.
2. The hamburger icon and the navigation bar background should have a clear visual distinction for better usability.

**Proposed Changes**:
To address these issues, we should implement the following changes:
1. Add a feature to close the mobile menu when clicking anywhere outside the menu area.
2. Adjust the styling so that the hamburger icon and the background color are visually distinct, making the icon more noticeable.

**Additional Information**:
- Current behavior is not user-friendly and can lead to frustration.
- The proposed changes will improve the mobile user experience and the overall aesthetics of the website.
- This enhancement is crucial for a more user-friendly and polished dashboard.

**Related Screenshots**:
![image](https://github.com/Real-Dev-Squad/website-dashboard/assets/70854507/53e92130-5bc9-4698-b92f-6f7064fe5f4f)

shritis98:
Can you please assign this task to me

NIKHILNAIR21:
I am interested to do this task, can you assign this task to me

iamitprakash:
Proposed solution doc link https://docs.google.com/document/d/1amGzFwhll1JvqQLAH9xy3EMY4HyQknabVCjTCSwrUYY/edit?pli=1

PR #571, opened 2023-10-14
fix/mobile-navbar: fixed issues in hamburger menu

shritis98:
### Issue:
Enhance Mobile Navbar Behavior and Styling #531

### Description:
The hamburger menu should close when the user clicks anywhere on the screen, outside of the navigation menu. And the hamburger menu and the page background should have a clear visual distinction for better usability.

### Feature flag for closing hamburger menu upon clicking anywhere on the screen:

https://github.com/Real-Dev-Squad/website-dashboard/assets/140648846/73857b6b-1500-4c37-9a69-6e200d5a2f83

### Implementation of CSS for the hamburger menu:

**Before**

image

**After**

image

### Test:

image

### DIFF
--- a/script.js
+++ b/script.js
@@ -312,6 +312,7 @@ addClickEventListener(
 
 const hamburgerDiv = document.querySelector('.hamburger');
 const navLinks = document.querySelector('.links');
+let navActive = document.querySelector('.nav-links');
 let toggle = true;
 
 hamburgerDiv.addEventListener('click', function () {
@@ -323,6 +324,15 @@ hamburgerDiv.addEventListener('click', function () {
     toggle = true;
   }
 });
+
+if (params.get('dev') === 'true') {
+  document.addEventListener('click', function (event) {
+    if (!navActive.contains(event.target)) {
+      navLinks.classList.remove('active');
+      toggle = true;
+    }
+  });
+}
 
 addClickEventListener(
   syncIdleUsersButton,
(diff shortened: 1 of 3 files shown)


script.js
```diff
@@ -324,6 +325,15 @@ hamburgerDiv.addEventListener('click', function () {
   }
 });
 
+if (params.get('dev') === 'true') {
+  document.addEventListener('click', function (event) {
```
iamitprakash, line 329: what If the user clicks on the navigation links, the navLinks.classList.remove('active'); line of code will be executed, which will remove the active class from all of the navigation links. This means that the user will not be able to tell which navigation link is currently active.

shritis98: For the dashboard site, navLinks refers to the class 'links' (hamburger menu) and class 'active' is added only to the class 'links'(hamburger menu), so it won't affect any other navbar links when 'active' gets removed, also there are no links added for the options in hamburger menu, so clicking on any of them opens dashboard site.

iamitprakash: > For the dashboard site, navLinks refers to the class 'links' (hamburger menu) and class 'active' is added only to the class 'links'(hamburger menu), so it won't affect any other navbar links when 'active' gets removed, also there are no links added for the options in hamburger menu, so clicking on any of them opens dashboard site.

cool, what about below code

document.addEventListener('click', function (event) {
  if (!navActive.contains(event.target) && !event.target.closest('.nav-link')) {
    navLinks.classList.remove('active');
    toggle = true;
  }
});

shritis98: !navActive.contains(event.target)
-here the constant 'navActive'  contains the class '.nav-links', so it will automatically check if the clicked element is a descendent of '.nav-links' or not.

iamitprakash (approved):

Issue #707, opened 2024-03-04
Search field for User's


rahul-singh04:
### Issue Description
Add search user field on Discord User's page

### Expected Behaviour
User should be able to search for a desired user using it's name.

### Current Behavior
No Such Functionality exists. New Feature addition. 

### Severity/Priority
 - [ ] Critical
 - [ ] High
 - [ ] Medium
 - [ ] Low

### Checklist

- [ ] I have read and followed the project's code of conduct.
- [ ] I have searched for similar issues before creating this one.
- [ ] I have provided all the necessary information to understand and reproduce the issue.
- [ ] I am willing to contribute to the resolution of this issue.

PR #716, opened 2024-03-14
Feature/search user

rahul-singh04:
Date: 14/03/2024

Developer Name: Rahul Singh

---

## Issue Ticket Number
Issue #707 

## Description

Added search user functionality for discord users.

### Documentation Updated?

- [ ] Yes
- [x] No

N/A
### Under Feature Flag

- [ ] Yes
- [x] No

N/A
### Database Changes

- [ ] Yes
- [x] No

N/A
### Breaking Changes

- [ ] Yes
- [x] No

### Development Tested?

- [X] Yes
- [ ] No

## Screenshots
<details>

Screenshot 2024-03-14 at 5 05 28 PM

</details>

## Test Coverage
<details>

Screenshot 2024-03-14 at 5 06 45 PM

</details>

All but one test case passes which has an issue of date due time zone difference. Would raise a seperate Issue to fix it.


__tests__/users/App.test.js
```diff
@@ -89,4 +89,53 @@ describe('App Component', () => {
...
+    const initialUrl = await page.url();
+
+    await page.waitForSelector('.search_field');
+
+    await page.type('.search_field', 'John Doe');
+    await page.click('.search_button');
+
+    const updatedUrl = await page.url();
+
+    expect(updatedUrl).not.toEqual(initialUrl);
```
prakashchoudhary07, line 103: instead of chacking for inequality, please check for the correctness
If the string got appended or not in the URL

I the current case if the URL becomes, realdevsquad.com also then tests will pass

__tests__/users/App.test.js
```diff
@@ -89,4 +89,53 @@ describe('App Component', () => {
...
+    const userDetailsDisplayed =
+      (await page.$('.user_details_section')) !== null;
+
+    expect(userDetailsDisplayed).toBeTruthy();
+  });
+
+  it('should display search results matching the search term', async () => {
+    await page.type('.search_field', 'amit');
+    await page.click('.search_button');
+
```
prakashchoudhary07, line 122: Question: how is a search being done here in testing, 
when the name is sent, does it make an API call, yes then where?

users/discord/App.js
```diff
@@ -46,23 +49,45 @@ const handleUserSelected = (e) => {
...
 };
+const users = usersData[activeTab] ?? [];
 
-export const App = () => {
-  const users = usersData[activeTab] ?? [];
+const filteredUsers = users.filter(
+  (user) =>
+    user.github_display_name &&
+    user.github_display_name.toLowerCase().includes(searchTerm.toLowerCase()),
+);
```
prakashchoudhary07, line 62: What is the search key?

prakashchoudhary07 (changes requested):

__tests__/users/App.test.js
rahul-singh04: fixed and pushed commit for this.

users/discord/App.js
rahul-singh04: on button click the search term gets appended in url and page Rerenders and the search term is fetched from there and then users are  filtered on basis of that search term.

__tests__/users/App.test.js
rahul-singh04: No, its not making an api call. But i changed the logic for test case now. After performing search it now checks if the page contains a result containing the search term.

users/discord/App.js
prakashchoudhary07: my question is based on which field is the search being performed here.
Is it based on a username? first name, last name, GitHub username?
Which field is used for the search?

rahul-singh04: Based on github displayname. Which for all users were there actual names.

Issue #738, opened 2024-04-06
[Bug] Message display inconsistency in extension requests page


mehra-sourav:
### Issue Description
The extension requests page does not display the empty page message after approving all extension requests, leaving a blank page instead.

### Expected Behavior
After approving all extension requests, the page should display the empty page message.

![image](https://github.com/Real-Dev-Squad/website-dashboard/assets/36883648/71efb682-ae1e-4639-8b83-67ce2d82bf40)

### Current Behavior
After approving all extension requests, the page becomes blank., which can cause confusion for users.

![image](https://github.com/Real-Dev-Squad/website-dashboard/assets/36883648/2359c211-b1ef-4c9c-8f24-887de65bd6ee)

### Screenshots

https://github.com/Real-Dev-Squad/website-dashboard/assets/36883648/5e1a505b-464f-4ec5-8b94-72a48c016bc6

### Reproducibility

 - [x] This issue is reproducible
 - [ ] This issue is not reproducible

### Steps to Reproduce
1. Sign in as super user
2. Approve/Reject any extension request

### Severity/Priority
 - [ ] Critical
 - [ ] High
 - [x] Medium
 - [ ] Low

### Checklist

- [x] I have read and followed the project's code of conduct.
- [x] I have searched for similar issues before creating this one.
- [x] I have provided all the necessary information to understand and reproduce the issue.
- [x] I am willing to contribute to the resolution of this issue.

PR #748, opened 2024-04-13
test: Display 'No extension requests to show!' message after addressing all requests

mehra-sourav:
Date: 13th April 2024

Developer Name: @mehra-sourav 

---

## Issue Ticket Number
#738 

## Description
This PR adds the tests of issue #738 .

### Documentation Updated?

- [ ] Yes
- [x] No

### Under Feature Flag

- [ ] Yes
- [x] No

### Database Changes

- [ ] Yes
- [x] No

### Breaking Changes

- [ ] Yes
- [x] No

### Development Tested?

- [x] Yes
- [ ] No

## Test Coverage
<details>

image

</details>


__tests__/extension-requests/extension-requests.test.js
```diff
@@ -358,6 +358,25 @@ describe('Tests the Extension Requests Screen', () => {
           },
           body: JSON.stringify(extensionRequestsListUserSearch),
         });
+      } else if (
+        url ===
+        'https://api.realdevsquad.com/extension-requests?order=asc&size=1&q=status%3APENDING'
```
iamitprakash, line 363: why this api url is hardcoded

__tests__/extension-requests/extension-requests.test.js
```diff
@@ -358,6 +358,25 @@ describe('Tests the Extension Requests Screen', () => {
           },
           body: JSON.stringify(extensionRequestsListUserSearch),
         });
+      } else if (
```
iamitprakash, line 361: potential issues 

imited Data: The response only provides the first pending extension request. If you need all pending requests, you'd modify body to include the entire extensionRequestsListPending.allExtensionRequests array.
Error Handling: The code doesn't handle potential errors (e.g., extensionRequestsListPending being empty). You might want to add checks and appropriate responses for such cases.

iamitprakash (changes requested):

__tests__/extension-requests/extension-requests.test.js
mehra-sourav: This is because for my test, I'm fetching that specific request's mock data from extensionRequestsListPending which is mocked to be successfully approved/denied. Some of the other requests, in extensionRequestsListPending are mocked to fail.

__tests__/extension-requests/extension-requests.test.js
mehra-sourav: 1. I have intentionally limited my fetching to that one specific mock data as it is the only one whose status can be successfully changed. Others in extensionRequestsListPending are mocked to fail.
2. The extensionRequestsListPending is mock data, so it shouldn't be empty, right? Please correct me if I'm understanding your question incorrectly.

__tests__/extension-requests/extension-requests.test.js
satyam73, line 1: I think the chain of else if and copying configs of request in every intercepting request can be improved by adding common configs in object constants.
But I think it can be picked later and ticket can be created for it.
Same applies to base urls since only the path and query params is changing in every intercepting request.

mehra-sourav: Sure, I'll create an issue for it.

__tests__/extension-requests/extension-requests.test.js
mamtakumari2305: > This is because for my test, I'm fetching that specific request's mock data from extensionRequestsListPending which is mocked to be successfully approved/denied. Some of the other requests, in extensionRequestsListPending are mocked to fail.

reason seems not correct to me, please confirm from @prakashchoudhary07

__tests__/extension-requests/extension-requests.test.js
mehra-sourav: I've added a null check for the case when `extensionRequestsListPending` is empty.

__tests__/extension-requests/extension-requests.test.js
```diff
@@ -919,4 +938,31 @@ describe('Tests the Extension Requests Screen', () => {
...
+    await page.waitForNetworkIdle();
+
+    extensionRequestsElement = await page.$('.extension-requests');
+    let extensionCardsList = await page.$$('.extension-card');
+
+    for (const card of extensionCardsList) {
+      let approveButton = await card.$('.approve-button');
+      await approveButton.click();
+      await page.waitForTimeout(2500);
+    }
```
prakashchoudhary07, line 953: When clicking on the approve button, what happens (Does it make API call to rds backend or is the behaviour intercepted)

prakashchoudhary07: If it's intercepted, is there a need to wait for 2500 seconds here?

If not, then how does that work?

__tests__/extension-requests/extension-requests.test.js
```diff
@@ -919,4 +938,31 @@ describe('Tests the Extension Requests Screen', () => {
     );
     expect(ascSortIconDisplayStyle).toBe('block');
   });
+
+  it('Should show empty message if all extension requests have been addressed', async () => {
```
prakashchoudhary07, line 942: Is there a check, if the API returns no data?

__tests__/extension-requests/extension-requests.test.js
mehra-sourav: The button triggers an API call, but we're intercepting it in the tests. However, after the API calls succeeds, an animation is triggered that lasts around 1600 ms. I've reduced the timeout to 1700 ms from 2500 ms.

__tests__/extension-requests/extension-requests.test.js
mehra-sourav: This test works even if no data is returned from the API. Is an explicit check still required?

prakashchoudhary07: but this test doesn't check for the empty case

__tests__/extension-requests/extension-requests.test.js
```diff
@@ -919,4 +939,44 @@ describe('Tests the Extension Requests Screen', () => {
...
+    const extensionRequestContainerText = await page.evaluate(
+      (element) => element.innerText,
+      extensionRequestsElement,
+    );
+
+    expect(extensionCardsList.length).toBe(0);
+    expect(extensionRequestContainerText).toBe(
+      'No extension requests to show!',
+    );
+  });
```
prakashchoudhary07, line 981: Which check will be tested here?

first one or the second one, how does it works?

mehra-sourav: Both checks will be tested. 

The first one will be checked when the API fetches 0 extension requests (no extension requests remaining).

The second one will be checked when the API fetches 1 or more extension requests.

__tests__/extension-requests/extension-requests.test.js
```diff
@@ -372,9 +372,11 @@ describe('Tests the Extension Requests Screen', () => {
...
           body: JSON.stringify({
             message: 'Extension Requests returned successfully!',
-            allExtensionRequests: [
-              extensionRequestsListPending?.allExtensionRequests?.[1] ?? {},
-            ],
+            allExtensionRequests:
+              extensionRequestsListPending?.allExtensionRequests.length > 0
+                ? [extensionRequestsListPending?.allExtensionRequests?.[1]]
+                : [],
+            // [],
```
iamitprakash: remove this

iamitprakash (changes requested):

iamitprakash (approved):

Issue #833, opened 2024-09-18
Enhancement:Implement Infinite Scroll Loading For Users


prashant67690:
### Issue Description

The User Management page currently features the Click Option through next button to load the other Users in the client Side, the goal of the task to rebuilt the way we load the  users on this page also the users cards have to be now rendered in the form of a list in the list view column currently they are not rendered as a list.  

### Expected Behavior

The Expected Behavior of the application after this enhancement feature will be that the Users will be now rendered as a List in List View and there will a infinite scroll bar for loading more users as we enter the end of the page. 
### Current Behavior

Currently The Users Are not Loading automatically when we hit the end of the page and Also the users are not rendered in the list view as a list. 
### Screenshots

![image](https://github.com/user-attachments/assets/dc9c6bb1-6006-45a6-bd53-7f9a0ac54bb6)

### Reproducibility

 - [x] This issue is reproducible
 - [ ] This issue is not reproducible

### Steps to Reproduce
1. go to https://dashboard.realdevsquad.com/users/ and scroll till bottom.
2. You will see the next and previous button and when you visit the page you will observe the non list rendering of the users 

### Severity/Priority
 - [ ] Critical
 - [ ] High
 - [x] Medium
 - [ ] Low

### Checklist

- [x] I have read and followed the project's code of conduct.
- [x] I have searched for similar issues before creating this one.
- [x] I have provided all the necessary information to understand and reproduce the issue.
- [x] I am willing to contribute to the resolution of this issue.

PR #860, opened 2024-09-28
Feature/Infinite-scrolling-feature-and-user-search-fix-on-user-management-page

prashant67690:
Date: 28-09-2024

Developer Name: Prashant Kumar 

---

## Issue Ticket Number

closes : [#833](https://github.com/Real-Dev-Squad/website-dashboard/issues/833)  
## Description
The User Management page currently features infinite scroll loading of users and the users in the list view are now rendered as the list rather than in the form of cards / tiles , due these changes the search functionality was having issue , that thing is fixed and running as expected down below.

### Documentation Updated?

- [ ] Yes
- [x] No

### Under Feature Flag

- [x] Yes
- [ ] No

### Database Changes

- [ ] Yes
- [x] No

### Breaking Changes

- [ ] Yes
- [x] No

### Development Tested?

- [x] Yes
- [ ] No

## Screenshots
<details>

Before:
prev
oldlist
371653915-742a5a61-42d0-48d9-a1a1-058c21169ad7

</details>
<details>
<summary>Screenshot 2</summary>

After:
loader
afterlist

</details>

## Test Coverage
<details>

tests
</details>

## Additional Notes

here we can see the search functionality behavior after the fixes 

https://github.com/user-attachments/assets/847504bb-caa7-4589-8f03-d7cd69437f03


vinit717 (changes requested):

users/script.js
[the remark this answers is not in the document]
prashant67690: which functionality , for checking weather user loader is happening or not is done , i just fixed some other thing that broke due to my change in the logic of loading more users

users/script.js
```diff
@@ -162,10 +160,18 @@ function generateUserList(
     return;
   }
 
-  const ulElement = document.getElementById('head_list');
+  console.log(users);
```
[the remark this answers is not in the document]
prashant67690: fixed this

users/script.js
```diff
@@ -162,10 +160,17 @@ function generateUserList(
     return;
   }
 
-  const ulElement = document.getElementById('head_list');
+  const errorTexts = document.getElementsByClassName('error-text');
+  Array.from(errorTexts).forEach((errorText) => {
+    errorText.remove();
+  });
```
yesyash, line 166: - do not select the element using classname use id instead. 
- as we know there will be only one element that will contain the error message do we need to create an array and loop over all elements that match the classname?

yesyash (changes requested):

users/script.js
```diff
@@ -162,10 +160,17 @@ function generateUserList(
...
+
+  if (showPagination || page == 0) {
+    ulElement.innerHTML = '';
+  }
 
-  if (users != null) {
-    users.forEach((userData) => {
+  if (users?.length !== 0) {
```
yesyash, line 172: - do we need an explicit `!== 0` check here?

users/script.js
```diff
@@ -31,21 +32,24 @@ const init = (
...
         return;
       }
-      run = false;
-      showUserDataList(
-        page++,
-        userListElement,
-        paginationElement,
-        loaderElement,
-      );
+      if (page === 0) {
```
yesyash, line 42: - do we need an explicit `===0` check here?

users/script.js
```diff
@@ -31,21 +32,24 @@ const init = (
...
+        page++;
```
yesyash, line 43: - can we do an early return here?

yesyash (changes requested):

prashant67690:
yash sir fixed the previous problems and also wrote the additional test as you requested please review

users/style.css
```diff
@@ -145,10 +149,21 @@ li p {
...
+  height: 50px;
+  animation: spin 2s linear infinite;
+  margin: 150px auto;
+}
+
 .pagination {
   display: flex;
   justify-content: space-between;
   margin-top: auto;
+  visibility: hidden;
```
Ayushsanjdev, line 166: You know that this element will still take the space if we use visibility hidden? Is it the use case? can you explain? Why not display: none

Ayushsanjdev (changes requested):

users/style.css
prashant67690: did this so other features that might depend on it are not broken and i can develop the required feature and there were some services that depend on it

Ayushsanjdev (approved):
LGTM

yesyash (approved):

vinit717 (approved):

PR #933, opened 2025-01-04
fix: update development server command

JC-Coder:
Date:  04/01/2025

Developer Name:  JC-Coder

---

## Description
update development server command to run the start script and initialize the proxy also to improve development flow

### Documentation Updated?

- [ ] Yes
- [ ] No

### Under Feature Flag

- [ ] Yes
- [x] No

### Database Changes

- [ ] Yes
- [x] No

### Breaking Changes

- [ ] Yes
- [x] No

### Development Tested?

- [x] Yes
- [ ] No

## Screenshots
<details>

image

https://github.com/user-attachments/assets/a4d35a52-1fc7-40f7-bfd9-129bcc63ae0a

</details>

## Test Coverage
<details>

</details>

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,7 +8,7 @@
     "fix": "prettier --write .",
     "test": "jest",
     "start": "node server.js",
-    "dev": "local-ssl-proxy --source 443 --target 5500 -n dev.realdevsquad.com"
+    "dev": "node server.js & local-ssl-proxy --source 443 --target 8000 -n dev.realdevsquad.com"
   },
   "repository": {
     "type": "git",


package.json
```diff
@@ -8,7 +8,7 @@
     "fix": "prettier --write .",
     "test": "jest",
     "start": "node server.js",
-    "dev": "local-ssl-proxy --source 443 --target 5500 -n dev.realdevsquad.com"
+    "dev": "node server.js & local-ssl-proxy --source 443 --target 8000 -n dev.realdevsquad.com"
```
MehulKChaudhari, line 11: Thanks for working on this. 

Suggestion: Can we use `node server.js && local-ssl-proxy --source 443 --target 8000 -n dev.realdevsquad.com` command instead? 

adding two `&&` will ensure local-ssl-proxy runs only if node server.js starts successfully this will avoid the issues where the proxy starts but the server fails. I think it will be a more reliable

yesyash (dismissed):

package.json
yesyash, line 11: - couple of questions:
   - what does `node server.js` do?
   - why are we changing the port from 5500 to 8000?

vinit717:
Also please add a working screen recording proof

package.json
JC-Coder: i'm not sure i understand why all this question. i basically made a change to combine 2 script instead of have user run 1 and get lost (mostly new user).

JC-Coder:
> Also please add a working screen recording proof

screenshot added . kindly check

package.json
vikhyat187: Its basically to run the server in background, if we use `&` with any command, its basically to create a background process @MehulKChaudhari .

vinit717:
> > Also please add a working screen recording proof
> 
> screenshot added . kindly check

not the screenshot can you please add a screen recording of the working website at dev and localhost

JC-Coder:
> > > Also please add a working screen recording proof
> > 
> > 
> > screenshot added . kindly check
> 
> not the screenshot can you please add a screen recording of the working website at dev and localhost

video added.

Issue #864, opened 2024-09-30
Create Superuser Dashboard for Managing Onboarding Extensions


ksubramanyeshwara:
### Issue Description
Design a user interface that allows moderators to efficiently manage extension requests, ensuring a smooth workflow and preventing unnecessary user removal due to incomplete assignments.

### Expected Behavior
Dashboard: Moderators have an intuitive interface to manage extension requests. 
The dashboard should display:
- User's name/ID
- Reason for request
- Requested number of days
- Approve and Deny buttons
- Moderators can adjust the number of days if necessary.
- Option to ask for reevaluation: Moderators should be able to request more information

### Current Behavior
There is no existing system or interface for requesting extensions, which leads to users being automatically removed after the 31-day onboarding period without a chance to extend their time. Moderators lack tools to manage onboarding timelines.

### Screenshots
Not applicable at the moment.

### Reproducibility

- [ ] This issue is reproducible
- [x] This issue is not reproducible

### Steps to Reproduce
N/A (Not applicable for feature implementation)

### Severity/Priority
 - [ ] Critical
 - [ ] High
 - [x] Medium
 - [ ] Low

#### UX Requirements

**User interface should display:**

- User name/ID
- Reason for request
- Requested number of extension days

**User interface for moderators :**

- Approve / Deny button
- Modify requested days
- Request reevaluation of days and reason

**Technical Requirements:**

- Integrate with existing user management system
- Provide notifications for users

### Checklist

- [x] I have read and followed the project's code of conduct.
- [x] I have searched for similar issues before creating this one.
- [x] I have provided all the necessary information to understand and reproduce the issue.
- [x] I am willing to contribute to the resolution of this issue.

PR #955, opened 2025-02-27
feat :  Implemented Logic for Filtering Requests by State and Username

AnujChhikara:
Date: 27 Feb, 2025

Developer Name: @AnujChhikara 

---

## Issue Ticket Number

- #864 
## Description
This PR adds the logic to filter requests based on state and username. The UI for filtering was already in place, but the filtering functionality was not implemented. With this update:

- Users can now filter requests by state (e.g., pending, approved, rejected).
- Users can also filter requests by username, allowing them to quickly find relevant requests.
- Debouncing implemented: When a user types a username, we wait 300ms before sending the request to minimize API calls and improve performance.

### Documentation Updated?

- [ ] Yes
- [x] No

### Under Feature Flag

- [x] Yes
- [ ] No

### Database Changes

- [ ] Yes
- [x] No

### Breaking Changes

- [ ] Yes
- [x] No

### Development Tested?

- [x] Yes
- [ ] No

## Screenshots
<details>

 - Filter by State

https://github.com/user-attachments/assets/c5153b17-e937-450e-b147-ee3d30af17a6

 - Filter by Username

https://github.com/user-attachments/assets/37c687de-12d9-490c-8056-7d7d0f15d0e1

 - Approving and rejecting Requests

https://github.com/user-attachments/assets/486b161e-ceae-4b3d-913a-604449ab1d75

</details>

## Test Coverage
<details>

image

</details>

## Additional Notes
- This is the last PR for this task 

## Summary by CodeRabbit

- **New Features**
  - Introduced a new user suggestion feature based on the search term "randhir".
  - Improved request filtering to display only approved requests when the filter is applied.
  - Enhanced assignee search with a debounced input, providing timely and relevant user suggestions.

- **Style**
  - Refined the appearance of suggestion elements with adjusted spacing and font size for better readability.

### DIFF
--- a/__tests__/requests/requests.test.js
+++ b/__tests__/requests/requests.test.js
@@ -6,7 +6,7 @@ const {
   extensionRequest,
   onboardingExtensionRequest,
 } = require('../../mock-data/requests');
-const { allUsersData } = require('../../mock-data/users');
+const { allUsersData, userRandhir } = require('../../mock-data/users');
 const {
   STAGING_API_URL,
   LOCAL_TEST_PAGE_URL,
@@ -105,6 +105,17 @@ describe('Tests the request cards', () => {
             'Access-Control-Allow-Headers': 'Content-Type, Authorization',
           },
           body: JSON.stringify(approvedRequest),
+        });
+      } else if (url === `${STAGING_API_URL}/users?search=randhir&size=5`) {
+        interceptedRequest.respond({
+          status: 200,
+          contentType: 'application/json',
+          headers: {
+            'Access-Control-Allow-Origin': '*',
+            'Access-Control-Allow-Methods': 'GET, POST, PUT, DELETE, OPTIONS',
+            'Access-Control-Allow-Headers': 'Content-Type, Authorization',
+          },
+          body: JSON.stringify(userRandhir),
         });
       } else {
         interceptedRequest.continue();
@@ -369,6 +380,52 @@ describe('Tests the request cards', () => {
         'input[type="radio"][name="status-filter"]:checked',
       );
       expect(selectedStatusRadio).toBeNull();
+    });
+
+    it('should render the correct requests based on the selected filter state', async () => {
+      const filterToggleButton = await page.$(
+        '[data-testid="filter-toggle-button"]',
+      );
+      await filterToggleButton.click();
+
+      const approvedRadio = await page.$(
+        'input[type="radio"][name="status-filter"][value="APPROVED"]',
+      );
+      await approvedRadio.click();
+
+      const applyFilterButton = await page.$(
+        '[data-testid="apply-filter-button"]',
+      );
+      await applyFilterButton.click();
+
+      const requestCards = await page.$$('[data-testid="ooo-request-card"]');
+
+      for (const card of requestCards) {
+        const statusText = await card.$eval(
+          '[data-testid="request-status"]',
+          (el) => el.textContent,
+        );
+        expect(statusText).toContain('Approved');
+      }
+    });
+
+    it('should show user suggestions after debounced input', async () => {
+      await page.waitForSelector('#assignee-search-input', { visible: true });
+      let username = 'randhir';
+      await page.type('#assignee-search-input', username, { delay: 100 });
+
+      await page.waitForNetworkIdle();
+
+      const suggestionCount = await page.$$eval(
+        '.suggestion',
+        (elements) => elements.length,
+      );
+      const suggestions = await page.$$eval('.suggestion', (elements) =>
+        elements.map((el) => el.textContent?.trim()),
+      );
+
+      expect(suggestionCount).toBeGreaterThan(0);
+      expect(suggestions).toContain(username);
     });
   });
 });
--- a/__tests__/requests/requests.test.js
+++ b/__tests__/requests/requests.test.js
@@ -409,8 +409,33 @@ describe('Tests the request cards', () => {
       }
     });
 
+    it('should show suggestions only after minimum character threshold', async () => {
+      await page.waitForSelector('#assignee-search-input', { visible: true });
+
+      const partialQuery = 'ra';
+      await page.type('#assignee-search-input', partialQuery, { delay: 100 });
+
+      await page.waitForNetworkIdle();
+
+      let suggestionCount = await page.$$eval(
+        '.suggestion',
+        (elements) => elements.length,
+      );
+
+      expect(suggestionCount).toBe(0);
+
+      await page.type('#assignee-search-input', 'ndhir', { delay: 100 });
+      await page.waitForNetworkIdle();
+      suggestionCount = await page.$$eval(
+        '.suggestion',
+        (elements) => elements.length,
+      );
+
+      expect(suggestionCount).toBeGreaterThan(0);
+    });
     it('should show user suggestions after debounced input', async () => {
       await page.waitForSelector('#assignee-search-input', { visible: true });
+      await page.$eval('#assignee-search-input', (input) => (input.value = ''));
       let username = 'randhir';
       await page.type('#assignee-search-input', username, { delay: 100 });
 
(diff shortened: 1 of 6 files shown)


requests/util.js
```diff
@@ -43,6 +43,9 @@ function getQueryParamsString(requestType, query) {
   ) {
     queryParam += `&state=${query.state}`;
   }
+  if (query.requestedBy !== undefined && query.requestedBy !== null) {
+    queryParam += `&requestedBy=${query.requestedBy}`;
+  }
```
MayankBansal12, line 48:
```suggestion
  if (query?.requestedBy != null) {
    queryParam += `&requestedBy=${query.requestedBy}`;
  }
```

mock-data/requests/index.js
```diff
@@ -20,20 +20,18 @@ const pendingRequest = {
...
+    createdAt: 1711439903761,
+    requestedBy: 'V4rqL1aDecNGoa1IxiCu',
+    from: 1712275200000,
+    until: 1712448000000,
+    type: 'OOO',
+    message: 'request message',
+    lastModifiedBy: 'V4rqL1aDecNGoa1IxiCu',
+    state: 'APPROVED',
+    updatedAt: 1711482912686,
+  },
```
MayankBansal12, line 34: will it not return an array? if not, then why have next and prev cursor?

AnujChhikara: This mock data is used to fetch a specific request that is being approved or rejected. The API ensures that when a superuser approves or rejects a request, we fetch only that particular request and update it accordingly. This is the [API](https://api.realdevsquad.com/requests?dev=true&id=QjLMT6j1jsAdlr8WWa6B) example.

MayankBansal12: Understood. The mock data has prev and next cursor along with data, i guess that can be removed now!

MayankBansal12 (approved):

RishiChaubey31 (dismissed):

requests/script.js
```diff
@@ -626,7 +643,92 @@ filterButton.addEventListener('click', (event) => {
...
+      if (users && users.length > 0) {
+        users.forEach((user) => {
+          const userDiv = document.createElement('div');
+          userDiv.className = 'suggestion';
+          userDiv.textContent = user.username;
+
+          suggestionContainer.appendChild(userDiv);
+
+          userDiv.addEventListener('click', async () => {
+            userNameFilterInput.value = user.username;
```
Achintya-Chatterjee, line 693: We should optimize debounce logic to prevent redundant API calls.
 - The current implementation clears the timeout but still allows multiple requests if the user types quickly.  
 - This can cause unnecessary load on the server and degrade performance.  
- Consider using _.debounce() from lodash or write a custom debounce function with a trailing execution strategy.  
- This ensures the API call is made only after the user stops typing, improving efficiency.  
- Additionally, caching previous results for the same input can further reduce API requests.

requests/script.js
```diff
@@ -626,7 +643,92 @@ filterButton.addEventListener('click', (event) => {
...
+
+            const requestData = { requestedBy: user.username };
+            await renderRequestCards(requestData);
+          });
+        });
+      }
+    }, 300);
+  } else {
+    suggestionContainer.classList.add('hidden');
+    suggestionContainer.innerHTML = '';
```
Achintya-Chatterjee, line 705: Clearing and repopulating `suggestionContainer (suggestionContainer.innerHTML = '';`) is an **anti-pattern**. Instead, We should use `DocumentFragment` to append elements in one go.
Also, if users re-type the same input, you're re-fetching data instead of caching previous results

requests/util.js
```diff
@@ -43,6 +43,9 @@ function getQueryParamsString(requestType, query) {
   ) {
     queryParam += `&state=${query.state}`;
   }
+  if (query?.requestedBy != null) {
+    queryParam += `&requestedBy=${query.requestedBy}`;
+  }
```
Achintya-Chatterjee, line 48: The query parameters are concatenated directly (`queryParam += '&requestedBy=' + query.requestedBy;`). If `queryParam` is empty, the query string starts with `&`instead of `?`. So we should use `new URLSearchParams(query).toString()` instead of manually appending query parameters.

requests/script.js
```diff
@@ -540,30 +552,35 @@ async function performAcceptRejectAction(isAccepted, e) {
       state: isAccepted ? 'APPROVED' : 'REJECTED',
     });
   }
+
   const parentDiv = e.target.closest('.request__card');
-  parentDiv.classList.add('disabled');
-  const removeSpinner = addSpinner(parentDiv);
+  requestPageContainer.classList.add('disabled');
+  const removeSpinner = addSpinner(requestPageContainer);
```
Achintya-Chatterjee, line 558: You are disabling the entire `requestPageContainer (requestPageContainer.classList.add('disabled')`) instead of only disabling the specific card. So this means that one action blocks all other interactions, which is a poor UX decision. Instead, we should only disable the clicked request card.

requests/util.js
```diff
@@ -151,3 +154,26 @@ function addRadioButton(labelText, value, groupName) {
...
+      headers: {
+        'Content-type': 'application/json',
+      },
+    });
+    const data = await res.json();
+    return data.users;
+  } catch (error) {
+    console.log(error);
+  }
+}
```
Achintya-Chatterjee, line 179: If the server returns a non-200 response, you are not checking `res.ok`, which may lead to unexpected behaviours.

requests/script.js
```diff
@@ -105,9 +111,15 @@ function updateUrlWithQuery(type) {
...
-    (nextLink || '/requests' + getQueryParamsString(requestType, query));
+  let finalUrl;
+  if (query?.state?.[0] || query?.requestedBy) {
+    finalUrl =
+      API_BASE_URL + '/requests' + getQueryParamsString(requestType, query);
+  } else {
+    finalUrl =
+      API_BASE_URL +
+      (nextLink || '/requests' + getQueryParamsString(requestType, query));
+  }
```
Achintya-Chatterjee, line 122: `getRequests()` function has redundant conditions. `if (query?.state?.[0] || query?.requestedBy) {...}` should be refactored. Why are we doing repetitive URL assignments, this is leading to unnecessary complexity.

requests/script.js
```diff
@@ -626,7 +643,92 @@ filterButton.addEventListener('click', (event) => {
...
+  if (!selectedState) {
+    requestContainer.innerHTML = '';
+    await renderRequestCards(requestData);
+  } else {
+    requestContainer.innerHTML = '';
+    requestData.state = selectedState;
+    await renderRequestCards(requestData);
+  }
+});
```
Achintya-Chatterjee, line 667: The function calls `renderRequestCards()` twice, once for `requestData.state === null` and once for `state !== null`. Instead of doing this, we should structure the logic cleanly.

Achintya-Chatterjee (changes requested):

requests/util.js
AnujChhikara: added that check

requests/script.js
AnujChhikara: Refactored this function

requests/script.js
AnujChhikara: Improved this, now only that card will get disabled when we perform action on that request card

requests/script.js
```diff
@@ -626,7 +641,94 @@ filterButton.addEventListener('click', (event) => {
...
+        suggestionContainer.appendChild(fragment);
+      }
+    }, 300);
+  } else {
+    suggestionContainer.classList.add('hidden');
+    while (suggestionContainer.firstChild) {
+      suggestionContainer.removeChild(suggestionContainer.firstChild);
+    }
+  }
+});
```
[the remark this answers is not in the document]
AnujChhikara: The getUsersByUsername function already has error handling that logs issues and returns an empty array on failure, so the app won’t crash.

AnujChhikara:
Hi @Achintya-Chatterjee, thanks for the review! I’ve addressed most of your comments. Regarding your suggestion to cache API calls, we currently don’t have a caching mechanism on the dashboard site. There are several options—like using a Map, a custom function, or other approaches—but I’d love your input on how to choose the best one for our use case. How do you think we should decide?

requests/script.js
Achintya-Chatterjee: we should write a custom debounce function, I think this will be the most sought-after approach.

AnujChhikara: Okay, we already have a custom function for debounce , I have used that one.

Achintya-Chatterjee: Cool

__tests__/requests/requests.test.js
```diff
@@ -370,5 +381,51 @@ describe('Tests the request cards', () => {
...
+      for (const card of requestCards) {
+        const statusText = await card.$eval(
+          '[data-testid="request-status"]',
+          (el) => el.textContent,
+        );
+        expect(statusText).toContain('Approved');
+      }
+    });
+
+    it('should show user suggestions after debounced input', async () => {
```
Achintya-Chatterjee, line 412: the test correctly verifies that suggestions appear and include the expected username, it would be beneficial to add assertions for edge cases such as empty results, special characters, or minimum character threshold.

Achintya-Chatterjee (changes requested):

__tests__/requests/requests.test.js
AnujChhikara: added please check

Achintya-Chatterjee (approved):

iamitprakash (approved):

PR #961, opened 2025-03-14
Fix: Incorrect Error Message and UI issues for Non-Super Users on Request Actions  

### LINKED ISSUE #959: Non-Super User Action on Requests Shows Wrong Error and Removes Cards
### Issue Description
When a non-super user clicks the "Accept" or "Reject" button on a requests, the error message displayed is incorrect. Additionally, while the error is correctly shown via a toast notification, all request cards disappear from the page. Instead of this behavior, the request cards should remain unchanged, and only the toast notification should display the error message.

### Expected Behavior
If a non-super user attempts to accept or reject a request:
- An appropriate error message should be displayed via toast notification only.
- The request cards should remain visible and unchanged.

### Current Behavior
- An incorrect error message is displayed.
- The error is shown via toast notification (this is correct).
- All request cards disappear from the page when the error occurs.
### Screenshots

Image

### Reproducibility

 - [X] This issue is reproducible
 - [ ] This issue is not reproducible
### Steps to Reproduce
- Navigate to this [Page](https://dashboard.realdevsquad.com/requests/?dev=true) 
- If you are not a super user try to click on accept or reject button
- you can notice the error 
### Severity/Priority
 - [ ] Critical
 - [ ] High
 - [x] Medium
 - [ ] Low

### Checklist

- [x] I have read and followed the project's code of conduct.
- [x] I have searched for similar issues before creating this one.
- [x] I have provided all the necessary information to understand and reproduce the issue.
- [x] I am willing to contribute to the resolution of this issue.

mridxl:
Date: 14 Mar 2025

Developer Name: @Mridxl

---

## Issue Ticket Number

- Closes #959 

## Description

This PR fixes the incorrect error message shown when a non-super user clicks the "Accept" or "Reject" button on requests. Additionally, it ensures that request cards remain visible on the page instead of disappearing when an error occurs. The error message will now be displayed only via a toast notification without altering the state of the request list.

### Documentation Updated?

- [ ] Yes
- [x] No

### Under Feature Flag

- [x] Yes
- [ ] No

### Database Changes

- [ ] Yes
- [x] No

### Breaking Changes

- [ ] Yes
- [x] No

### Development Tested?

- [x] Yes
- [ ] No

## Screenshots
<details>

image

</details>

<details>
<summary>Screen Recording</summary>

https://github.com/user-attachments/assets/cd806716-be43-4e4e-a9c4-66fb9c67d12f

</details>

## Test Coverage
<details>

image

</details>

### DIFF
--- a/requests/script.js
+++ b/requests/script.js
@@ -511,8 +511,8 @@ async function acceptRejectRequest(id, reqBody) {
     } else {
       switch (res.status) {
         case 401:
-          showToast(ErrorMessages.UNAUTHENTICATED, 'failure');
-          showMessage('ERROR', ErrorMessages.UNAUTHENTICATED);
+          showToast(ErrorMessages.UNAUTHORIZED_ACTION, 'failure');
+          showMessage('ERROR', ErrorMessages.UNAUTHORIZED_ACTION);
           break;
         case 403:
           showToast(ErrorMessages.UNAUTHENTICATED, 'failure');
@@ -585,6 +585,9 @@ async function performAcceptRejectAction(isAccepted, e) {
       showMessage('ERROR', ErrorMessages.SERVER_ERROR);
     }
   }
+
+  nextLink = '';
+  await renderRequestCards({ state: statusValue, sort: sortByValue });
 }
 
 function showToast(message, type) {
--- a/requests/constants.js
+++ b/requests/constants.js
@@ -27,6 +27,7 @@ const ErrorMessages = {
   OOO_NOT_FOUND: 'OOO Requests not found',
   EXTENSION_NOT_FOUND: 'Extension Requests not found',
   ONBOARDING_EXTENSION_NOT_FOUND: 'Onboarding extension Requests not found',
+  UNAUTHORIZED_ACTION: 'You are unauthorized to perform this action',
   SERVER_ERROR: 'Unexpected error occurred',
 };
 


requests/script.js
```diff
@@ -585,6 +585,9 @@ async function performAcceptRejectAction(isAccepted, e) {
       showMessage('ERROR', ErrorMessages.SERVER_ERROR);
     }
   }
+
+  nextLink = '';
```
MayankBansal12, line 589: where is this `nextLink` being used?

mridxl: nextLink is used to track pagination. (along with the intersectionObserver set up)

Resetting nextLink to an empty string ensures than when `created requests` are re-fetched, we receive the most recent requests. Not resetting it fetches the next page (` GET /requests` with nextLink set to the last request's id), and we receive older requests instead of the most recent ones.

mridxl:
> Could you please add a video in the description for better clarity? 
Yes, I've added the video in the description.

> Also, is it appropriate to display the error message in both the toast and the body?
The error message was being displayed in the body earlier, so I assumed that was how we wanted it to be.
IMO, we should only display a toast to notify the user and not display the error in the body.

AnujChhikara:
> > Could you please add a video in the description for better clarity?
> 
> Yes, I've added the video in the description.
> 
> > Also, is it appropriate to display the error message in both the toast and the body?
> 
> The error message was being displayed in the body earlier, so I assumed that was how we wanted it to be. IMO, we should only display a toast to notify the user and not display the error in the body.

A toast seems sufficient for notifying the user—maybe we can check with others for their input. Also, since we're re-rendering the cards, is there a way to prevent that when an error occurs to avoid an extra DB call?

Achintya-Chatterjee (approved):

RishiChaubey31 (approved):

PR #989, opened 2025-04-29
Feature Reusable filter component implementation

AnujChhikara:
Date: 29 April 2025

Developer Name: @AnujChhikara 

---

## Issue Ticket Number

- #987 
## Description

This PR adds the implementation of the reusable Filter component.

- The component has been integrated into the following pages:

```
/applications

/extension-requests

/requests

/task-requests
```

Features
- The Filter component becomes visible when dev=true is added to the URL query.

- Users can filter content based on status or state.

- On applying a filter, a tag is displayed to indicate the active filter—enhancing the overall user experience.

- The selected filter updates the URL query parameters, and the component also reads from the URL to reflect the correct filter state on page load.

Related PRs
- Test PR #990 
- Function PR #988
### Documentation Updated?

- [ ] Yes
- [x] No

### Under Feature Flag

- [x] Yes
- [ ] No

### Database Changes

- [ ] Yes
- [x] No

### Breaking Changes

- [ ] Yes
- [x] No

### Development Tested?

- [x] Yes
- [ ] No

## Screenshots
<details>
<summary>Screenshot 1</summary>

- Application Page

https://github.com/user-attachments/assets/59bfa0a0-f707-4b47-9fac-67fe7d871a06

- Extension Requests Page

https://github.com/user-attachments/assets/641fa769-f72a-49b4-b3f5-39e0d5ccb591

- Requests Page

https://github.com/user-attachments/assets/b4983e32-77ef-485d-a1aa-948aa4ebc9f5

- Task-Requests Page

https://github.com/user-attachments/assets/402058f7-537a-4c5c-8b95-27174c05f494

</details>

## Test Coverage
<details>
<summary>Screenshot 1</summary>

image
image

image
image
image

</details>

## Additional Notes
- We previously had similar tag functionality on the task-requests page, which was guarded by a dev flag. Some tests were already written for that. However, with the new filter component implementation, we plan to centralize most of these tests into a single file instead of duplicating them across multiple files. That’s why I haven’t renamed the existing tests — once the dev flag is removed, those tests will be cleaned up as well.

PR #988, opened 2025-04-28
Feature: Add utility function for rendering filter component

AnujChhikara:
Date: 28 April 2025

Developer Name: @AnujChhikara 

---

## Issue Ticket Number

 - #987
## Description

This PR adds the util function that will be used to render the filter component

- This function will be used into the following pages:

```
/applications

/extension-requests

/requests

/task-requests
```

Related PRs
- Test PR #990 
- Implementation PR #989
### Documentation Updated?

- [ ] Yes
- [x] No

### Under Feature Flag

- [x] Yes
- [ ] No

### Database Changes

- [ ] Yes
- [x] No

### Breaking Changes

- [ ] Yes
- [x] No

### Development Tested?

- [x] Yes
- [ ] No

## Screenshots
<details>
<summary>Screenshot 1</summary>

- Application Page

https://github.com/user-attachments/assets/59bfa0a0-f707-4b47-9fac-67fe7d871a06

- Extension Requests Page

https://github.com/user-attachments/assets/641fa769-f72a-49b4-b3f5-39e0d5ccb591

- Requests Page

https://github.com/user-attachments/assets/b4983e32-77ef-485d-a1aa-948aa4ebc9f5

- Task-Requests Page

https://github.com/user-attachments/assets/402058f7-537a-4c5c-8b95-27174c05f494

</details>

## Test Coverage
<details>
<summary>Screenshot 1</summary>

image
image

image
image
image
</details>

Issue #987, opened 2025-04-28
Create Reusable Components for Filter Component


AnujChhikara:
### Issue Description
- We currently have different filter components across multiple pages, leading to inconsistencies in design and behavior.
- To improve reusability, maintainability, and consistency across the application, we need to create a single reusable filter component that can be used across different pages. This filter component should be flexible enough to handle dynamic filter options and various use cases.

### Expected Behavior
- A single, reusable filter component that can be easily used on different pages.

- It should follow the application's existing design system for consistency

### Current Behavior
- Filter components are implemented differently across various pages.

- Lack of standardization makes maintenance and updates more difficult.

### Screenshots

- filter component on different pages

![Image](https://github.com/user-attachments/assets/3cf4d1fd-26d1-41ad-9930-59062dfdfec0)

![Image](https://github.com/user-attachments/assets/90e67d5a-3b70-4e0d-acc8-32471c71103f)

![Image](https://github.com/user-attachments/assets/dce216e5-6440-4b12-9dda-e125ec7a7db2)

### Reproducibility

 - [X] This issue is reproducible
 - [ ] This issue is not reproducible

### Severity/Priority
 - [ ] Critical
 - [ ] High
 - [x] Medium
 - [ ] Low

### Checklist

- [x] I have read and followed the project's code of conduct.
- [x] I have searched for similar issues before creating this one.
- [x] I have provided all the necessary information to understand and reproduce the issue.
- [x] I am willing to contribute to the resolution of this issue.

AnujChhikara:
Closing this issue as it has been resolved in PR  #1000

PR #990, opened 2025-04-29
Add Tests for Reusable Filter Component

AnujChhikara:
Date: 29 March 2025

Developer Name: @AnujChhikara 

---

## Issue Ticket Number

- #987 
## Description

This PR adds test coverage for the reusable Filter component.

- A dedicated test file has been created to validate the core functionality of the Filter component, including UI rendering and button event listeners.

- Since the Filter component is used across four different pages, we've also added tests in each respective page's test file to ensure that the component renders correctly and displays the appropriate content in each context.

Feature PRs

- #988 
- #989

### Documentation Updated?

- [ ] Yes
- [x] No

### Under Feature Flag

- [x] Yes
- [ ] No

### Database Changes

- [ ] Yes
- [x] No

### Breaking Changes

- [ ] Yes
- [x] No

### Development Tested?

- [x] Yes
- [ ] No

## Screenshots
<details>

- Application Page

https://github.com/user-attachments/assets/59bfa0a0-f707-4b47-9fac-67fe7d871a06

- Extension Requests Page

https://github.com/user-attachments/assets/641fa769-f72a-49b4-b3f5-39e0d5ccb591

- Requests Page

https://github.com/user-attachments/assets/b4983e32-77ef-485d-a1aa-948aa4ebc9f5

- Task-Requests Page

https://github.com/user-attachments/assets/402058f7-537a-4c5c-8b95-27174c05f494

</details>

## Test Coverage
<details>

image
image

image
image
image

</details>

### DIFF
--- a/__tests__/applications/applications.test.js
+++ b/__tests__/applications/applications.test.js
@@ -361,4 +361,20 @@ describe('Applications page', () => {
       'application updated successfully!',
     );
   });
+
+  it('should display only accepted applications after applying the "ACCEPTED" filter when dev=true', async () => {
+    await page.goto(`${LOCAL_TEST_PAGE_URL}/applications?dev=true`);
+    await page.waitForNetworkIdle();
+    await page.click('[data-testid="filter-component-toggle-button"]');
+    const applyFilterButton = '[data-testid="apply-filter-component-button"]';
+    await page.waitForSelector(applyFilterButton, { visible: true });
+
+    await page.click(`input[type="checkbox"][id="ACCEPTED"]`);
+
+    await page.click(applyFilterButton);
+
+    await page.waitForNetworkIdle();
+    const applicationCardElements = await page.$$('.application-card');
+    expect(applicationCardElements.length).toBe(acceptedApplications.length);
+  });
 });
--- a/__tests__/applications/applications.test.js
+++ b/__tests__/applications/applications.test.js
@@ -130,9 +130,6 @@ describe('Applications page', () => {
 
   afterEach(async () => {
     await page.close();
-  });
-
-  afterAll(async () => {
     await browser.close();
   });
 
@@ -368,11 +365,8 @@ describe('Applications page', () => {
     await page.click('[data-testid="filter-component-toggle-button"]');
     const applyFilterButton = '[data-testid="apply-filter-component-button"]';
     await page.waitForSelector(applyFilterButton, { visible: true });
-
     await page.click(`input[type="checkbox"][id="ACCEPTED"]`);
-
     await page.click(applyFilterButton);
-
     await page.waitForNetworkIdle();
     const applicationCardElements = await page.$$('.application-card');
     expect(applicationCardElements.length).toBe(acceptedApplications.length);
--- a/__tests__/components/filter.test.js
+++ b/__tests__/components/filter.test.js
@@ -1,0 +1,266 @@
+const puppeteer = require('puppeteer');
+
+const {
+  fetchedApplications,
+  acceptedApplications,
+} = require('../../mock-data/applications');
+const { superUserForAudiLogs } = require('../../mock-data/users');
+const {
+  STAGING_API_URL,
+  LOCAL_TEST_PAGE_URL,
+} = require('../../mock-data/constants');
+
+describe('Filter Component Core Functionality', () => {
+  let browser;
+  let page;
+
+  jest.setTimeout(60000);
+
+  const filterButtonSelector = '[data-testid="filter-component-toggle-button"]';
+  const filterModalSelector = '[data-testid="filter-component-modal"]';
+  const applyFilterButtonSelector =
+    '[data-testid="apply-filter-component-button"]';
+  const clearFilterButtonSelector =
+    '[data-testid="filter-component-clear-button"]';
+  const activeFilterTagsSelector = '[data-testid="active-filter-tags"]';
+
+  beforeEach(async () => {
+    browser = await puppeteer.launch({
+      headless: 'new',
+      ignoreHTTPSErrors: true,
+      args: ['--incognito', '--disable-web-security'],
+    });
+    page = await browser.newPage();
+    await page.setRequestInterception(true);
+
+    page.on('request', (interceptedRequest) => {
+      const url = interceptedRequest.url();
+      if (url === `${STAGING_API_URL}/applications?dev=true`) {
+        interceptedRequest.respond({
+          status: 200,
+          contentType: 'application/json',
+          body: JSON.stringify({
+            applications: acceptedApplications,
+            totalCount: acceptedApplications.length,
+          }),
+          headers: {
+            'Access-Control-Allow-Origin': '*',
+            'Access-Control-Allow-Methods': 'GET, POST, PUT, DELETE, OPTIONS',
+            'Access-Control-Allow-Headers': 'Content-Type, Authorization',
+          },
+        });
+      } else {
+        interceptedRequest.continue();
+      }
+    });
+
+    await page.goto(`${LOCAL_TEST_PAGE_URL}/applications?dev=true`);
+    await page.waitForSelector(filterButtonSelector);
+  });
+
+  afterEach(async () => {
+    await page.close();
+  });
+  afterAll(async () => {
+    await browser.close();
+  });
+
+  it('should render filter toggle button and hide filter modal when page load', async () => {
+    const filterButton = await page.$(filterButtonSelector);
+    expect(filterButton).toBeTruthy();
+    const filterModal = await page.$(filterModalSelector);
+    expect(filterModal).toBeTruthy();
+    const isModalHidden = await page.$eval(filterModalSelector, (el) =>
+      el.classList.contains('hidden'),
+    );
+    expect(isModalHidden).toBe(true);
+  });
+
+  it('should open filter modal with options on when click on filter toggle button', async () => {
+    await page.click(filterButtonSelector);
+    await page.waitForSelector(`${filterModalSelector}:not(.hidden)`);
+
+    const isModalHidden = await page.$eval(filterModalSelector, (el) =>
+      el.classList.contains('hidden'),
+    );
+    expect(isModalHidden).toBe(false);
+    const applyButton = await page.$(applyFilterButtonSelector);
+    expect(applyButton).toBeTruthy();
+    const clearButton = await page.$(clearFilterButtonSelector);
+    expect(clearButton).toBeTruthy();
+    const checkbox = await page.$(`input[type="checkbox"][id="ACCEPTED"]`);
+    expect(checkbox).toBeTruthy();
+  });
+
+  it('should toggle the clear button state based on checkbox selection', async () => {
+    await page.click(filterButtonSelector);
+    await page.waitForSelector(`${filterModalSelector}:not(.hidden)`);
+
+    let isClearButtonDisabled = await page.$eval(
+      clearFilterButtonSelector,
+      (el) => el.disabled,
+    );
+    expect(isClearButtonDisabled).toBe(true);
+
+    await page.click(`input[type="checkbox"][id="ACCEPTED"]`);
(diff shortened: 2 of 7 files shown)


__tests__/task-requests/task-request.test.js
```diff
@@ -42,7 +42,9 @@ describe('Task Requests', () => {
         });
       } else if (
         url ===
-        `${STAGING_API_URL}/taskRequests?size=20&q=status%3Aapproved++sort%3Acreated-asc`
+          `${STAGING_API_URL}/taskRequests?size=20&q=status%3Aapproved++sort%3Acreated-asc` ||
+        url ===
+          `${STAGING_API_URL}/taskRequests?size=20&q=status%3Aapproved+sort%3Acreated-asc`
```
AnujChhikara, line 47: The reason for adding this URL here is because it's called when approving a task request. Even though both URLs look almost identical, there's a small difference — one uses a single + before sort, and the other uses two +. So we included both to handle that minor variation and avoid any issues.

__tests__/extension-requests/extension-requests.test.js
```diff
@@ -1209,4 +1209,26 @@ describe('Tests the Extension Requests Screen', () => {
       'Extension request successfully updated.',
     );
   });
+
+  it('should show only "APPROVED" and "DENIED" extension requests after selecting those filters when dev=true', async () => {
```
Suvidh-kaushik, line 1213: if this test checks for only APPROVED and DENIED why are we checking all extensionRequests.length at the end won't it also contrain PENDING requests

Suvidh-kaushik (changes requested):

__tests__/extension-requests/extension-requests.test.js
AnujChhikara: So when that API is called to get the approved and pending requests, it responds with the mock extensionRequests data. That's why checking extensionRequests.length

Suvidh-kaushik: but the test is to show only approved and denied requests and won't the extensionRequests.length you are checking have all types of requests?

Suvidh-kaushik: but the test is to show only approved and denied requests and won't the extensionRequests.length you are checking have all types of requests?

__tests__/applications/applications.test.js
```diff
@@ -361,4 +361,20 @@ describe('Applications page', () => {
...
+    const applyFilterButton = '[data-testid="apply-filter-component-button"]';
+    await page.waitForSelector(applyFilterButton, { visible: true });
+
+    await page.click(`input[type="checkbox"][id="ACCEPTED"]`);
+
+    await page.click(applyFilterButton);
+
+    await page.waitForNetworkIdle();
+    const applicationCardElements = await page.$$('.application-card');
+    expect(applicationCardElements.length).toBe(acceptedApplications.length);
```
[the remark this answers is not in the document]
AnujChhikara: yes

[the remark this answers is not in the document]
ZendeAditya: Okay

__tests__/components/filter.test.js
```diff
@@ -0,0 +1,266 @@
...
+    const isModalHidden = await page.$eval(filterModalSelector, (el) =>
+      el.classList.contains('hidden'),
+    );
+    expect(isModalHidden).toBe(false);
+    const applyButton = await page.$(applyFilterButtonSelector);
+    expect(applyButton).toBeTruthy();
+    const clearButton = await page.$(clearFilterButtonSelector);
+    expect(clearButton).toBeTruthy();
+    const checkbox = await page.$(`input[type="checkbox"][id="ACCEPTED"]`);
+    expect(checkbox).toBeTruthy();
```
ZendeAditya, line 92: Shouldn't we display all three options in the modal instead of just `ACCEPTED` when the user opens it?

ZendeAditya: Or is this because we are on the `/application` page? As you write in the code, do we only show `ACCEPTED/APPROVED` on the `application` page?

__tests__/components/filter.test.js
```diff
@@ -0,0 +1,266 @@
...
+  it('should apply filter, update URL, and display the corresponding filter tag', async () => {
+    const statusToApply = 'ACCEPTED';
+
+    await page.click(filterButtonSelector);
+    await page.waitForSelector(`${filterModalSelector}:not(.hidden)`);
+
+    await page.click(`input[type="checkbox"][id="${statusToApply}"]`);
+
+    await page.click(applyFilterButtonSelector);
+    await page.waitForSelector(filterModalSelector + '.hidden');
```
ZendeAditya, line 129: Can't we write simply like `await page.waitForSelector(`${filterModalSelector}.hidden`);
`?? Instead of the concatenation of the string, which is a little bit difficult to understand a little bit?

__tests__/components/filter.test.js
```diff
@@ -0,0 +1,266 @@
...
+    await page.waitForNetworkIdle();
+
+    expect(page.url()).toContain(`status=${statusToApply.toLowerCase()}`);
+
+    const filterTagSelector = `${activeFilterTagsSelector} .filter__component__tag[data-status="${statusToApply}"]`;
```
ZendeAditya, line 134: Please do like this so it looks consistence.

__tests__/components/filter.test.js
```diff
@@ -0,0 +1,266 @@
...
+  });
+
+  it('should clear filter via Clear button', async () => {
+    const statusToApply = 'PENDING';
+
+    await page.click(filterButtonSelector);
+    await page.waitForSelector(`${filterModalSelector}:not(.hidden)`);
+    await page.click(`input[type="checkbox"][id="${statusToApply}"]`);
+    await page.click(applyFilterButtonSelector);
+    await page.waitForSelector(filterModalSelector + '.hidden');
```
ZendeAditya, line 151: if possible please update here also

__tests__/components/filter.test.js
```diff
@@ -0,0 +1,266 @@
...
+    await page.waitForNetworkIdle();
+    expect(page.url()).toContain(`status=${statusToApply.toLowerCase()}`);
```
ZendeAditya, line 153: After clearing the filter(Approved), the URL should not contain status=? filter?

AnujChhikara: @ZendeAditya where are we clearing the filter ??

__tests__/components/filter.test.js
AnujChhikara: fixed thank you for pointing it out

__tests__/components/filter.test.js
AnujChhikara: done

__tests__/components/filter.test.js
AnujChhikara: done

__tests__/requests/requests.test.js
```diff
@@ -585,4 +600,24 @@ describe('Tests the request cards', () => {
...
+    const applyFilterButton = '[data-testid="apply-filter-component-button"]';
+    await page.waitForSelector(applyFilterButton, { visible: true });
+    await page.click(`input[type="checkbox"][id="APPROVED"]`);
+    await page.click(applyFilterButton);
+    await page.waitForNetworkIdle();
+
+    const requestCards = await page.$$('[data-testid="ooo-request-card"]');
+    expect(requestCards.length).toBe(approvedRequestsData.data.length);
+
+    const statusText = await requestCards[0].$eval(
```
ZendeAditya, line 617: Hey Anuj, why the check for only the first card instead of all? May `ooo-request-card` other request card also?

__tests__/components/filter.test.js
```diff
@@ -0,0 +1,275 @@
...
+    const isClearButtonEnable = await page.$eval(
+      clearFilterButtonSelector,
+      (el) => !el.disabled,
+    );
+    expect(isClearButtonEnable).toBe(true);
+  });
+
+  it('should apply only one filter on application page', async () => {
+    await page.click(filterButtonSelector);
+    await page.waitForSelector(`${filterModalSelector}:not(.hidden)`);
```
ZendeAditya, line 252: The line `await page.waitForSelector(`${filterModalSelector}:not(.hidden)`);` appears more than ten times in the test file. Could you create a separate utility function for this? For example:

```javascript
const waitForFilterModal = async (page: Page, selector: string) => {
  await page.waitForSelector(`${selector}:not(.hidden)`);
};
```

Then, you can call this function each time with:

```javascript
await waitForFilterModal(page, filterModalSelector);
```

Thank you!

__tests__/requests/requests.test.js
AnujChhikara: We could do that, but it would be redundant since it would require looping through each card and checking them individually, which feels unnecessary in this case

__tests__/components/filter.test.js
```diff
@@ -0,0 +1,275 @@
...
+      `input[type="checkbox"][id="ACCEPTED"]`,
+      (el) => el.checked,
+    );
+    expect(isAcceptedChecked).toBe(true);
+
+    isPendingChecked = await page.$eval(
+      `input[type="checkbox"][id="PENDING"]`,
+      (el) => el.checked,
+    );
+    expect(isPendingChecked).toBe(false);
```
ZendeAditya, line 273: Why don't you add `Rejected`? If somebody selects `Reject`, then? add a block for check `Reject`
check

AnujChhikara: If we add a separate check for "Rejected", then we’d also need to add similar blocks for every other status. Since the logic remains the same for each scenario, adding multiple explicit checks would be unnecessary and repetitive

__tests__/requests/requests.test.js
ZendeAditya: But what if it shows other cards instead of `Approved`?

AnujChhikara: In line 615 we are already checking if the card we are rendering are equal to the api response of approved requests

ZendeAditya: Okay then there is no problem

__tests__/components/filter.test.js
```diff
@@ -0,0 +1,275 @@
...
+
+    const acceptedStatus = 'ACCEPTED';
+    const tag = await page.$(
+      `${activeFilterTagsSelector} .filter__component__tag[data-status="${acceptedStatus}"]`,
+    );
+    expect(tag).toBeTruthy();
+
+    await page.click(filterButtonSelector);
+    await page.waitForSelector(`${filterModalSelector}:not(.hidden)`);
+    const isCheckboxChecked = await page.$eval(
```
pankajjs, line 238:
```suggestion
    const isAcceptedChecked = await page.$eval(
```

__tests__/components/filter.test.js
```diff
@@ -0,0 +1,275 @@
...
+      `input[type="checkbox"][id="${acceptedStatus}"]`,
+      (el) => el.checked,
+    );
+    expect(isCheckboxChecked).toBe(true);
+    const isClearButtonEnable = await page.$eval(
```
pankajjs, line 243:
```suggestion
    const isClearButtonEnabled = await page.$eval(
```

pankajjs:
Please update the test screenshots

__tests__/task-requests/task-request.test.js
MayankBansal12: instead of using hardcoded urls, can we use `URLSearchParams` here?
it will help you avoid this situation

AnujChhikara: since the rest of the tests match full URLs as strings. Didn't want to overcomplicate for one edge case — can always refactor later if needed.

ZendeAditya (approved):

surajmaity1 (approved):
LGTM

iamitprakash (approved):

